PR #2, opened 2019-03-26
 Complement the ingest change that qualifies the account with its type

itzg:
# Resolves

Part of https://jira.rax.io/browse/SALUS-262

# What

This PR complements https://github.com/racker/salus-event-engine-ingest/pull/2 by resolving the Kapacitor database name with the same qualified account convention.

I also noticed the app context loading unit test was succeeding, but throwing errors about H2 not supporting the JSON column type used in the scenario entity. 

# How

Use a configuration driven approach that in production will include mappings like "hybrid -> CORE" and default to "ENCORE".

## How to test

Added a unit test for the tenant to qualified account mapping logic.

### DIFF
--- a/src/main/java/com/rackspace/salus/event/manage/config/TenantMappingProperties.java
+++ b/src/main/java/com/rackspace/salus/event/manage/config/TenantMappingProperties.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2019 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


itzg:
@jjbuchan I definitely agree; however, it's all to avoid hardcoding mentions of "hybrid" and "CORE" in this open source code base. As far as where it's used, it's to handle the potentially larger scope of ingested account types vs smaller scope of Identity/authentication tenants.

itzg:
To add to the note, the `LabelNamespaces` from [salus-telemetry-model](https://github.com/racker/salus-telemetry-model/pull/38/files#diff-333ce0b9f272ecee04b2df955791dc32) would be the other bit that can go into `umb-common`.

PR #4, opened 2019-05-01
added documentation to this project

harageth:
# Resolves

    SALUS-349

# What

    Adds internal and external documentation for Monitor-management

# How

    Generates internal documentation with Swagger then utilizes the SwaggerJsonConverter to export the internal documentation for external use

## How to test

    After Installing (Running the mvn install target) the SwaggerJsonConverter will be in the .m2 folder and then you can run the `test` target here that will generate the internal swagger and run the SwaggerJsonConverter

# TODO

    This should be complete

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -106,6 +106,18 @@
       <artifactId>springfox-swagger-ui</artifactId>
       <version>${springfox-swagger.version}</version>
     </dependency>
+
+    <dependency>
+      <groupId>io.swagger</groupId>
+      <artifactId>swagger-hibernate-validations</artifactId>
+      <version>1.5.6</version>
+    </dependency>
+    <dependency>
+      <groupId>com.rackspace.salus</groupId>
+      <artifactId>salus-tools</artifactId>
+      <version>0.0.1-SNAPSHOT</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -106,6 +106,18 @@
       <artifactId>springfox-swagger-ui</artifactId>
       <version>${springfox-swagger.version}</version>
     </dependency>
+
+    <dependency>
+      <groupId>io.swagger</groupId>
+      <artifactId>swagger-hibernate-validations</artifactId>
```
itzg: Same comment as
- https://github.com/racker/salus-telemetry-resource-management/pull/20#discussion_r279062003
- https://github.com/racker/salus-telemetry-monitor-management/pull/18#discussion_r279468770

src/main/java/com/rackspace/salus/event/manage/web/TasksApi.java
```diff
@@ -35,8 +37,17 @@
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 
+
 @RestController
-@RequestMapping("/api/tasks")
+@RequestMapping("/api/tasks/tenant")
```
itzg: I think you'll still find the conversion of paths for the public swagger won't line up to

https://github.com/racker/salus-telemetry-api/blob/master/public/src/main/java/com/rackspace/salus/telemetry/api/web/EventTasksController.java#L33
https://github.com/racker/salus-telemetry-api/blob/master/public/src/main/java/com/rackspace/salus/telemetry/api/web/EventTasksController.java#L52

I'm not sure if changing here to "/api/eventTasks/tenant" will be sufficient. You might need to enhance SwaggerJsonConverter to accept a replacement rule via command line.

Another option would be to "conform" with other backend services and change this request mapping to just "/api" and in each method's mapping it would be "/tenant/{tenantId}/eventTasks" etc

templates/markdown.hbs
```diff
@@ -0,0 +1,128 @@
+#{{#info}}{{title}}
+
+
+## {{join schemes " | "}}://{{host}}{{basePath}}
+
+
+{{description}}
+
+{{#contact}}
+[**Contact the developer**](mailto:{{email}})
```
itzg: Same comment as other PRs

src/main/java/com/rackspace/salus/event/manage/web/TasksApi.java
harageth: Yeah I hadn't actually looked at the path in the public API. I will change salus tools to accept CLI arguments for replacement rules.

PR #6, opened 2019-05-07
add labelSelector matching to kapacitor

harageth:
# Resolves

SALUS-193 https://jira.rax.io/browse/SALUS-193

# What

It adds labels to all of the relevant objects and then decomposes those labels into a comparison that kapacitor can accomplish

# How

It configures the TICKScript template to do the comparisons inside of kapacitor. 

## How to test

Right now you should import @itzg 's new insomnia profile. 
Start event-engine-management, event-engine-ingest, and bring up the docker components.
Then create a new Task Object from geoffs insomnia profile with
`"labelSelector": {"os": "linux"}` tacked onto the end of the "taskParameters" section of the JSON body.

# Why
Kapacitor has very limited capacity to do comparisons. We need to do the comparison in kapacitor so that if the labels get updated somewhere else then we automatically get the changes through the data pipeline. 

# TODO

The TICK script is not seeing the values that I am trying to tell it to reference. So that needs to be fixed


src/main/java/com/rackspace/salus/event/manage/entities/EventEngineTask.java
```diff
@@ -54,4 +53,12 @@
...
   @Column(nullable = false, columnDefinition = "json")
   TaskParameters taskParameters;
+
+  /**
+   * Labels are an indexed and query'able aspect of resources that are used for monitor
+   * matching.
+   */
+  @ElementCollection(fetch = FetchType.EAGER)
+  @CollectionTable(name="task_labels", joinColumns = @JoinColumn(name="id"))
+  Map<String,String> labels;
```
itzg: Call this `labelSelector` to be consistent.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -53,15 +57,35 @@ public TickScriptBuilder(TaskIdGenerator taskIdGenerator,
...
+    int first = 0;
+
+    Set<Map.Entry<String, String>> labels = taskParameters.getLabelSelector().entrySet();
+    for(Map.Entry<String, String> tuple: labels) {
+      if(first != 0) {
+        builder.append(" AND ");
+      }
+
+      //  This should get properly qualified in its own function
+      builder.append("\"system.systemMetadata." + tuple.getKey() + "\" == \"" + tuple.getValue()+"\"");
```
itzg: Ingest has to normalize several of the metrics object fields into tags, so instead of accessing via the `ExternalMetric` path you're using, you'll see at

https://github.com/racker/salus-event-engine-ingest/blob/master/src/main/java/com/rackspace/salus/event/ingest/services/IngestService.java#L104-L109

that it is applying a namespace (which resolves to prefix of `monitoring_system.metadata.`) to the label names.

harageth: Ahhhhh yeah I had run across that code and forgotten about it. As you can see from my comment I meant to properly qualify this but was just trying to make it work first so that I knew how to qualify it properly.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -53,15 +60,34 @@ public TickScriptBuilder(TaskIdGenerator taskIdGenerator,
 
   public String build(String tenantId, String measurement,  TaskParameters taskParameters) {
     return taskTemplate.execute(TaskContext.builder()
+            //how exactly do I want to build this? I could add it to the TaskObject. Or I could just create a function here...
```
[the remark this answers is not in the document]
harageth: Oh yeah... Forgot I added that. I usually try and add inner monologue when I start a task to clarify what and where I need to make my changes. Lets me clearly think about why I should do each different approach.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -53,15 +60,34 @@ public TickScriptBuilder(TaskIdGenerator taskIdGenerator,
...
+    StringBuilder builder = new StringBuilder();
+    int first = 0;
+
+    Set<Map.Entry<String, String>> labels = taskParameters.getLabelSelector().entrySet();
+    for(Map.Entry<String, String> tuple: labels) {
+      if(first != 0) {
+        builder.append(" AND ");
+      }
+
+      builder.append("if(isPresent(\""+LabelNamespaces.applyNamespace(MONITORING_SYSTEM_METADATA,tuple.getKey()) + "\"), \""+ LabelNamespaces.applyNamespace(MONITORING_SYSTEM_METADATA,tuple.getKey()) +"\" == '" + tuple.getValue()+"', FALSE )");
```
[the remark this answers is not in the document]
harageth: I was noticing that. I feel like this is similar to the Stored Procedure problem. I think I want @jjbuchan input on whether he cares before I move all of this into some mustache functions.

src/main/java/com/rackspace/salus/event/manage/model/CreateTask.java
```diff
@@ -21,6 +21,8 @@
 import javax.validation.constraints.NotNull;
 import lombok.Data;
 
+import java.util.Map;
```
jjbuchan:
```suggestion
```

src/main/java/com/rackspace/salus/event/manage/model/CreateTaskResponse.java
```diff
@@ -16,6 +16,7 @@
 
 package com.rackspace.salus.event.manage.model;
 
+import java.util.Map;
```
jjbuchan:
```suggestion
```

src/main/java/com/rackspace/salus/event/manage/model/EventEngineTaskDTO.java
```diff
@@ -17,6 +17,8 @@
 package com.rackspace.salus.event.manage.model;
 
 import com.rackspace.salus.event.manage.model.TaskParameters;
+
+import java.util.Map;
```
jjbuchan:
```suggestion
```

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -53,15 +60,44 @@ public TickScriptBuilder(TaskIdGenerator taskIdGenerator,
...
+    StringBuilder builder = new StringBuilder();
+    int first = 0;
+
+    Set<Map.Entry<String, String>> labels = taskParameters.getLabelSelector().entrySet();
+    for(Map.Entry<String, String> tuple: labels) {
+      if(first != 0) {
+        builder.append(" AND ");
+      }
+
+      builder.append(
```
itzg: I actually meant declare a variable like

```
String nsKey = LabelNamespaces.applyNamespace(MONITORING_SYSTEM_METADATA, key);
```

and then use `nsKey` instead of the duplicate call to apply namespace.

src/main/java/com/rackspace/salus/event/manage/entities/EventEngineTask.java
```diff
@@ -54,4 +53,12 @@
   @Type(type = "json")
   @Column(nullable = false, columnDefinition = "json")
   TaskParameters taskParameters;
+
+  /**
+   * Labels are an indexed and query'able aspect of resources that are used for monitor
```
itzg: nit: should be "event task matching"

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -53,15 +60,44 @@ public TickScriptBuilder(TaskIdGenerator taskIdGenerator,
...
+              "if(isPresent(\""+
+                      applyNamespace(tuple.getKey()) +
+                      "\"), \""+ //end of isPresent
+                      applyNamespace(tuple.getKey()) +
+                      "\" == '" +
+                      tuple.getValue()+
+                      "', FALSE )");
```
jjbuchan: What's the `FALSE` for?

jjbuchan: Your comments have gone from this section now, but I'm not sure what I think of using the mustache templating here... I'd have to see how it looks first.

But, I definitely agree that some reformatting could be used to make things easier to read.

I'm pretty sure when using `builder.append` you should not try to put everything into one call like you have here.  You can split it up into multiple calls and remove all the `+` concatenations.  I think if you use the `F2` tip I mentioned before, IntelliJ may actually say the same.

src/main/java/com/rackspace/salus/event/manage/model/CreateTask.java
jjbuchan: I'm guessing these suggestions will still leave a new empty line, so you still might have to change things locally if you commit the suggestion.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
itzg: Good point about `build.append`. Another option is to use `String.format` to improve readability.

I'm also thinking a good argument for doing more of this in the template is that you can avoid the quote escaping there.

src/main/java/com/rackspace/salus/event/manage/model/CreateTask.java
harageth: yeah probably. I accidentally committed these in my initial commit.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
harageth: Yeah that's definitely true of `builder.append()`. 

As for the mustache formatting it would look somewhat similar to this: https://github.com/racker/salus-event-engine-management/blob/master/templates/markdown.hbs#L86-L104

The FALSE is supposed to detect that the label doesn't exist and so the labels can't match, so it tells kapacitor that this comparison doesn't work. Unfortunately it doesn't appear to be preventing the error messages I was seeing.

src/test/java/com/rackspace/salus/event/manage/services/TickScriptBuilderTest.java
```diff
@@ -34,18 +41,25 @@
...
   @Test
-  public void testBuild() {
-    String expectedString = "stream\n  |from()\n    .measurement('measurement')\n" +
-            "    .groupBy('resourceId')\n  |alert()\n    .stateChangesOnly()\n" +
-            "    .id('tenant:{{index .Tags \"system.resourceId\"}}:measurement:field')\n" +
-            "    .details('task={{.TaskName}}')\n    .crit(lambda: \"field\" >= 33)\n    .topic('events')";
-
-    TaskParameters tp = new TaskParameters().setComparator(">=").setField("field").setThreshold(33);
+  public void testBuild() throws IOException{
+    String expectedString = readContent("/TICKScriptTemplateTest");
```
itzg: I prefer a convention where the test data file is somehow named relative to the test suite (TickScriptBuilderTest) and/or the test case (testBuild), so that it is intuitive what to name additional files. Also, give the test data file as suffix. It looks like `.tick` is the convention: https://docs.influxdata.com/kapacitor/v1.5/guides/load_directory/#tasks

src/test/java/com/rackspace/salus/event/manage/services/TickScriptBuilderTest.java
```diff
@@ -42,7 +42,7 @@
 
   @Test
   public void testBuild() throws IOException{
-    String expectedString = readContent("/TICKScriptTemplateTest");
+    String expectedString = readContent("/TICKScriptBuilderTest.tick");
```
itzg: Closer :) I meant `/TickScriptBuilderTest/testBuild.tick`

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
itzg: If the template approach doesn't work out, I'd like to see the `nsKey` code change I suggested.

harageth: I made that change on Thursday in case I had Jury Duty

https://github.com/racker/salus-event-engine-management/pull/6/files#diff-863e4661785ee1ff1aeebcbab27ec3beR80

itzg: Sorry, the PR diff led me astray.

src/main/resources/templates/task.mustache
```diff
@@ -2,7 +2,9 @@ stream
   |from()
     .measurement('{{measurement}}')
     .groupBy('{{groupBy}}')
-  |where(lambda: {{labelSelectorExpression}})
+{{#entrySet}}
+  |where(lambda: if(isPresent("{{labelNamespace}}.{{key}}"), "{{labelNamespace}}.{{key}}" == '{{value}}', FALSE))
```
itzg: Will multiple `where` nodes effectively `and` these conditions?

harageth: That's what I was just testing... It doesn't appear to process the metrics. The issue is that it implies the looping when a list is present and I was trying to not generate the where for when the entry set wasn't present without passing in a boolean value.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -71,6 +74,8 @@ public String build(String tenantId, String measurement,  TaskParameters taskPar
   @Data @Builder
   public static class TaskContext {
     Set<Map.Entry<String, String>> entrySet;
+    boolean entrySetIsSet;
```
itzg: Seeing this change now makes me think a better name than "entrySet" is needed for these two fields. How about simply "labels" and "labelsAvailable"?

itzg:
@harageth is kapacitor now working properly with the new template approach?

harageth:
> @harageth is kapacitor now working properly with the new template approach?

Yes

harageth:
But I forgot to update the test file to reflect the new changes.

I think I am going to add two more tests. One for the no-labels case and one for the multiple labels case.

src/test/java/com/rackspace/salus/event/manage/services/TickScriptBuilderTest.java
```diff
@@ -34,18 +41,25 @@
...
 
   @Test
-  public void testBuild() {
-    String expectedString = "stream\n  |from()\n    .measurement('measurement')\n" +
-            "    .groupBy('resourceId')\n  |alert()\n    .stateChangesOnly()\n" +
-            "    .id('tenant:{{index .Tags \"system.resourceId\"}}:measurement:field')\n" +
-            "    .details('task={{.TaskName}}')\n    .crit(lambda: \"field\" >= 33)\n    .topic('events')";
-
-    TaskParameters tp = new TaskParameters().setComparator(">=").setField("field").setThreshold(33);
+  public void testBuild() throws IOException{
```
itzg: It would be great to have just one more test case with more than one label tuple. Then the template "last" logic can be exercised.

src/main/java/com/rackspace/salus/event/manage/model/TaskParameters.java
```diff
@@ -29,4 +30,6 @@
   Number threshold;
   @NotEmpty
   String comparator;
+
+  Map<String, String> labelSelector;
```
itzg: Oh, you should use the new `@ValidLabelKeys` here

src/main/resources/templates/task.mustache
```diff
@@ -2,6 +2,18 @@ stream
   |from()
     .measurement('{{measurement}}')
     .groupBy('{{groupBy}}')
+{{#labelsAvailable}}
+  |where(lambda:
+  {{#labels}}
+    if(isPresent("{{labelNamespace}}.{{key}}"), "{{labelNamespace}}.{{key}}" == '{{value}}', FALSE)
```
itzg: Your other underscore-related change made me realize that event engine ingest is using `applyNamespace` here

https://github.com/racker/salus-event-engine-ingest/blob/master/src/main/java/com/rackspace/salus/event/ingest/services/IngestService.java#L107

so it'll be using the underscore delimiter between the metadata namespace and each label

https://github.com/racker/salus-telemetry-model/blob/master/src/main/java/com/rackspace/salus/telemetry/model/LabelNamespaces.java#L43

So I would expect this to now be `{{labelNamespace}}_{{key}}`. I'm then thinking that complexity of namespacing should be done up in the `build` method using `LabelNamespaces.applyNamespace` on each label entry's key.

harageth: I was expecting the same thing... But then I convinced myself that it should work as is. I could have sworn I saw that it was working correctly but now that I am looking at the results in kapacitor it isn't. There is something going on in the references for tags right now that I am trying to figure debug.

PR #8, opened 2019-05-21
Event-engine Consecutive Count 

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-192
https://jira.rax.io/browse/SALUS-366
https://jira.rax.io/browse/SALUS-365

367 is partially done. It needs testing added and probably needs the fields parameter converted into a list/array so that we can apply multiple fields in the one window
https://jira.rax.io/browse/SALUS-367
hasn't been started yet
https://jira.rax.io/browse/SALUS-368

# What

Adds consecutive count to the template for alarming and adds a couple of different levels of state change.

# How

It sets an expression for the stateCount function to track which allows us to reference that output for our alarms themselves.

# Why

Kapacitor has built out most of these features for us. So we are just using the chaining of Nodes that they have setup. 

## How to test

Currently you will need to start up the infrastructure, start up event-engine-ingest, start up event-engine-management

Create a new object like this 
`{
  "measurement": "cpu",
  "taskParameters": {
    "type": "Rising",
		"comparator": ">",
    "field": "used",
    "threshold": 1,
		"consecutiveCritical": 3, 
		"labelSelector": {
			"resource_metadata_os": "linux"
		}
  }
}`

Start a watch in Kapacitor on the object that was created
And then start the MetricsGenApplication

# TODO

Finish windowing
   Add in tests
   Make sure that we can pass many fields in at once and that it will grab all of them
Start and finish complex expressions


jjbuchan:
I assume I can ignore this until the other PR is merged.

harageth:
> I assume I can ignore this until the other PR is merged.

Yep. I'll make this run consecutive count based on time windows. (consecutively good for a certain amount of time is what Mitch wants).

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -77,17 +70,18 @@ public String build(String tenantId, String measurement, TaskParameters taskPara
...
         .warnCount(
           taskParameters.getWarning() != null ?
             String.format("\"warn_count\" >= %d",taskParameters.getWarning().getConsecutiveCount())
             : null)
         .critCount(
-            taskParameters.getCritical() != null ?
-                String.format("\"crit_count\" >= %d", taskParameters.getCritical().getConsecutiveCount())
-                :null)
+          taskParameters.getCritical() != null ?
+            String.format("\"crit_count\" >= %d", taskParameters.getCritical().getConsecutiveCount())
```
itzg: It would be nice to not force the "bake period" of 1 minute. Do you know if a count of 0 for these will effectively turn that off and have the alert fire on first match of the expression?

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -77,17 +70,18 @@ public String build(String tenantId, String measurement, TaskParameters taskPara
         .infoExpression(buildTICKExpression(taskParameters.getInfo()))
         .warnExpression(buildTICKExpression(taskParameters.getWarning()))
         .infoCount(
-            taskParameters.getInfo() != null ?
-                String.format("\"info_count\" >= %d", taskParameters.getInfo().getConsecutiveCount())
-                : null)
+          taskParameters.getInfo() != null ?
```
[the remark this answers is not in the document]
harageth: Yeah I was realizing that. Was mostly wanting to get tests created.

src/main/resources/templates/task.mustache
```diff
@@ -39,4 +42,8 @@ stream
 {{#critCount}}
     .crit(lambda: {{critCount}})
 {{/critCount}}
+{{#flapping}}
+    .flapping(.25, .5)
```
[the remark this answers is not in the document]
harageth: Yeah thats legitimate. I was realizing that as I was doing this that that might be a better way to do that.
I did add the history here to be explicit in the value that is supplied (its the default). I will do a similar thing with that.

src/test/java/com/rackspace/salus/event/manage/model/TaskParametersValidatorTest.java
```diff
@@ -0,0 +1,109 @@
...
+  public void setUp() {
+    validatorFactoryBean = new LocalValidatorFactoryBean();
+    validatorFactoryBean.afterPropertiesSet();
+  }
+
+  @Test
+  public void testValidation_normal_critical() {
+
+    final TaskParameters resource = new TaskParameters()
+        .setCritical(new TaskParameters().new LevelExpression()
```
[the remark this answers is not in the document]
harageth: Is this a suggestion to make it static? Doesn't static enforce a singleton which would mean every time I instantiate a new LevelExpression it overrides the other ones?

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
harageth: A value of 0 will trigger immediately

src/test/java/com/rackspace/salus/event/manage/model/TaskParametersValidatorTest.java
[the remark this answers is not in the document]
itzg: Yes, declare it as `static`. Java kinda overloads the keyword `static`...for inner classes it doesn't have anything to do with singleton-ness.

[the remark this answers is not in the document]
harageth: Oh perfect. I was avoiding it because of that but this will make things much easier.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -106,6 +105,10 @@ public String buildTICKExpression(LevelExpression expression) {
     String warnExpression;
     String infoExpression;
     @Default
+    float flappingLower = (float).25;
```
itzg: FYI you can also declare literal floats as `0.25f` with the trialing `f`

src/main/resources/templates/task.mustache
```diff
@@ -43,7 +54,7 @@ stream
     .crit(lambda: {{critCount}})
 {{/critCount}}
 {{#flapping}}
-    .flapping(.25, .5)
+    .flapping({{flappingLower}}, {{flappingUpper}})
 {{/flapping}}
     .history(21)
```
itzg: To finish out the values, moving this `history` up to task context would be good.

src/main/resources/templates/task.mustache
```diff
@@ -2,17 +2,17 @@ stream
...
-  |max({{field}})
-    .as('max_{{field}}')
-  |min({{field}})
-    .as('min_{{field}}')
-  |mean({{field}})
-    .as('avg_{{field}}')
-{{/windowing}}
+    .period({{windowLength}}m)
+    .every({{windowLength}}m)
+  |max({{windowField}})
```
itzg: Giving @GeorgeJahad a head start? :)

harageth: Like I said earlier. Windowing is mostly done. Although the windowField should ideally be a list or array of some sort. Then a max, min, avg can be created for each field provided.

PR #9, opened 2019-05-21
Allow different expressions for different levels for consecutiveCount

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-192 specifically the different levels of thresholds portion

# What

Customers might want different expressions for different levels of alarming. This is a proposal for altering the TaskParameters to store the expressions for each different level so that we can setup StateCount expressions for each of them. 

# How

It creates a new Expression object and then creates a number of levelExpressions that contain the consecutive count that correspond to the different levels that we expect. 

# Why

This is meant as a proposal. The objects encapsulate the functionality we need. It is inefficient in the sense that it doesn't allow for one base expression and then all of the sub expressions to use the consecutive count from that. 

It does however make the template easier. We can create new StateCount nodes for each level giving each one a different name with the `.as("warningLevel")` function which then allows us to create a single alarm and then add the lambdas that strictly reference those StateCount's. 

## How to test

This isn't even built into the template yet and shouldn't work because of data type changes that haven't been propogated to TickScriptBuilder yet.

# TODO
Update the tests

### DIFF
--- a/src/main/java/com/rackspace/salus/event/manage/model/Expression.java
+++ b/src/main/java/com/rackspace/salus/event/manage/model/Expression.java
@@ -1,10 +1,13 @@
 package com.rackspace.salus.event.manage.model;
 
+import com.rackspace.salus.event.manage.model.validator.ExpressionValidator;
+import com.rackspace.salus.event.manage.model.validator.ExpressionValidator.ComparatorValidation;
 import javax.validation.constraints.NotEmpty;
 import javax.validation.constraints.NotNull;
 import lombok.Data;
 
 @Data
+@ExpressionValidator.ComparatorValidation()
 public class Expression {
   @NotEmpty
   String field;
--- a/src/main/java/com/rackspace/salus/event/manage/model/Expression.java
+++ b/src/main/java/com/rackspace/salus/event/manage/model/Expression.java
@@ -7,13 +7,14 @@ import javax.validation.constraints.NotNull;
 import lombok.Data;
 
 @Data
-@ExpressionValidator.ComparatorValidation()
+
 public class Expression {
   @NotEmpty
   String field;
   @NotNull
   Number threshold;
   @NotEmpty
+  @ExpressionValidator.ComparatorValidation()
   String comparator;
 
 }
--- a/src/main/java/com/rackspace/salus/event/manage/model/Expression.java
+++ b/src/main/java/com/rackspace/salus/event/manage/model/Expression.java
@@ -1,0 +1,16 @@
+package com.rackspace.salus.event.manage.model;
+
+import javax.validation.constraints.NotEmpty;
+import javax.validation.constraints.NotNull;
+import lombok.Data;
+
+@Data
+public class Expression {
+  @NotEmpty
+  String field;
+  @NotNull
+  Number threshold;
+  @NotEmpty
+  String comparator;
+
+}
--- a/src/main/java/com/rackspace/salus/event/manage/model/TaskParameters.java
+++ b/src/main/java/com/rackspace/salus/event/manage/model/TaskParameters.java
@@ -32,7 +32,8 @@ public class TaskParameters {
   @ValidLabelKeys
   Map<String, String> labelSelector;
 
-  private class LevelExpression {
+  @Data
+  public class LevelExpression {
     Expression expression;
     Integer consecutiveCount;
   }
--- a/src/main/java/com/rackspace/salus/event/manage/model/TaskParameters.java
+++ b/src/main/java/com/rackspace/salus/event/manage/model/TaskParameters.java
@@ -25,15 +25,15 @@ import java.util.Map;
 
 @Data
 public class TaskParameters {
-  levelExpression consecutiveInfo;
-  levelExpression consecutiveWarning;
-  levelExpression consecutiveCritical;
+  LevelExpression info;
+  LevelExpression warning;
+  LevelExpression critical;
 
   @ValidLabelKeys
   Map<String, String> labelSelector;
 
-  private class levelExpression {
-    Expression baseExpression;
+  private class LevelExpression {
+    Expression expression;
     Integer consecutiveCount;
   }
 }
--- a/src/main/java/com/rackspace/salus/event/manage/model/TaskParameters.java
+++ b/src/main/java/com/rackspace/salus/event/manage/model/TaskParameters.java
@@ -25,17 +25,15 @@ import java.util.Map;
 
 @Data
 public class TaskParameters {
-  @NotEmpty
-  String field;
-  @NotNull
-  Number threshold;
-  @NotEmpty
-  String comparator;
-
-  Integer consecutiveInfo;
-  Integer consecutiveWarning;
-  Integer consecutiveCritical;
+  levelExpression consecutiveInfo;
+  levelExpression consecutiveWarning;
+  levelExpression consecutiveCritical;
 
   @ValidLabelKeys
   Map<String, String> labelSelector;
+
+  private class levelExpression {
+    Expression baseExpression;
+    Integer consecutiveCount;
+  }
 }
--- a/src/main/java/com/rackspace/salus/event/manage/model/TaskParameters.java
+++ b/src/main/java/com/rackspace/salus/event/manage/model/TaskParameters.java
@@ -16,6 +16,7 @@
 
 package com.rackspace.salus.event.manage.model;
 
+import com.rackspace.salus.event.manage.model.validator.TaskParametersValidator;
 import com.rackspace.salus.telemetry.model.ValidLabelKeys;
 import lombok.Data;
 
@@ -24,6 +25,7 @@ import javax.validation.constraints.NotNull;
 import java.util.Map;
 
 @Data
+@TaskParametersValidator.AtLeastOneOf()
 public class TaskParameters {
   LevelExpression info;
   LevelExpression warning;
--- a/src/main/java/com/rackspace/salus/event/manage/model/validator/ExpressionValidator.java
+++ b/src/main/java/com/rackspace/salus/event/manage/model/validator/ExpressionValidator.java
@@ -1,0 +1,39 @@
+package com.rackspace.salus.event.manage.model.validator;
+
+import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import com.rackspace.salus.event.manage.model.Expression;
+import com.rackspace.salus.event.manage.model.validator.TaskParametersValidator.AtLeastOneOf;
+import com.rackspace.salus.event.manage.types.Comparator;
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+import javax.validation.Constraint;
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import javax.validation.Payload;
+
+/**
+ * Validates on object creation whether the validator is one of the four acceptable options that we allow
+ */
+public class ExpressionValidator implements ConstraintValidator<AtLeastOneOf, Expression> {
+
+  @Override
+  public boolean isValid(Expression expression, ConstraintValidatorContext context) {
+    return Comparator.valid(expression.getComparator());
+  }
+
+  @Target({TYPE, ANNOTATION_TYPE}) // class level constraint
+  @Retention(RUNTIME)
+  @Constraint(validatedBy = ExpressionValidator.class) // validator
+  @Documented
+  public @interface ComparatorValidation {
+    String message() default "Valid comparators are: >, >=, <, <="; // default error message
+
+    Class<?>[] groups() default {}; // required
+
+    Class<? extends Payload>[] payload() default {}; // required
+  }
+}
--- a/src/main/java/com/rackspace/salus/event/manage/model/validator/ExpressionValidator.java
+++ b/src/main/java/com/rackspace/salus/event/manage/model/validator/ExpressionValidator.java
@@ -1,6 +1,7 @@
 package com.rackspace.salus.event.manage.model.validator;
 
 import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
+import static java.lang.annotation.ElementType.FIELD;
 import static java.lang.annotation.ElementType.TYPE;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
@@ -25,7 +26,7 @@ public class ExpressionValidator implements ConstraintValidator<AtLeastOneOf, Ex
     return Comparator.valid(expression.getComparator());
   }
 
-  @Target({TYPE, ANNOTATION_TYPE}) // class level constraint
+  @Target({FIELD, ANNOTATION_TYPE}) // class level constraint
   @Retention(RUNTIME)
   @Constraint(validatedBy = ExpressionValidator.class) // validator
   @Documented
--- a/src/main/java/com/rackspace/salus/event/manage/model/validator/ExpressionValidator.java
+++ b/src/main/java/com/rackspace/salus/event/manage/model/validator/ExpressionValidator.java
@@ -2,11 +2,9 @@ package com.rackspace.salus.event.manage.model.validator;
 
 import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
 import static java.lang.annotation.ElementType.FIELD;
-import static java.lang.annotation.ElementType.TYPE;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
-import com.rackspace.salus.event.manage.model.Expression;
-import com.rackspace.salus.event.manage.model.validator.TaskParametersValidator.AtLeastOneOf;
+import com.rackspace.salus.event.manage.model.validator.ExpressionValidator.ComparatorValidation;
 import com.rackspace.salus.event.manage.types.Comparator;
 import java.lang.annotation.Documented;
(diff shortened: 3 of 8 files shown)


src/main/java/com/rackspace/salus/event/manage/model/TaskParameters.java
```diff
@@ -25,17 +25,15 @@
...
-  Integer consecutiveWarning;
-  Integer consecutiveCritical;
+  levelExpression consecutiveInfo;
+  levelExpression consecutiveWarning;
+  levelExpression consecutiveCritical;
 
   @ValidLabelKeys
   Map<String, String> labelSelector;
+
+  private class levelExpression {
```
itzg: Java class names should start with uppercase.

src/main/java/com/rackspace/salus/event/manage/model/TaskParameters.java
```diff
@@ -25,17 +25,15 @@
...
-  String field;
-  @NotNull
-  Number threshold;
-  @NotEmpty
-  String comparator;
-
-  Integer consecutiveInfo;
-  Integer consecutiveWarning;
-  Integer consecutiveCritical;
+  levelExpression consecutiveInfo;
```
itzg: Since we need to support both expression-driven levels, such as "warning at 50%, critical at 90%", and consecutive counts then it seems like these fields should be simply `info`, `warning`, `critical`.

src/main/java/com/rackspace/salus/event/manage/model/TaskParameters.java
```diff
@@ -25,17 +25,15 @@
...
+  levelExpression consecutiveWarning;
+  levelExpression consecutiveCritical;
 
   @ValidLabelKeys
   Map<String, String> labelSelector;
+
+  private class levelExpression {
+    Expression baseExpression;
```
itzg: Thinking as an end user, `baseExpression` doesn't seem intuitive. How about just `expression`?

harageth: Yeah that seems more reasonable

src/main/java/com/rackspace/salus/event/manage/model/TaskParameters.java
harageth: Yeah they should. I just quickly altered it from what was there.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -61,22 +63,38 @@ public String build(String tenantId, String measurement,  TaskParameters taskPar
...
+        .alertId(taskIdGenerator.generateAlertId(tenantId, measurement, taskParameters.getCritical().getExpression().getField()))
         .labelsAvailable(labelsAvailable)
         .measurement(measurement)
         .details("task={{.TaskName}}")
-        .critExpression(String.format("\"%s\" %s %s", taskParameters.getField(), taskParameters.getComparator(), taskParameters.getThreshold()))
+        .critExpression(String.format("\"%s\" %s %s",
+            taskParameters.getCritical().getExpression().getField(),
+            taskParameters.getCritical().getExpression().getComparator(),
+            taskParameters.getCritical().getExpression().getThreshold()))
+        .infoExpression(taskParameters.getInfo() != null ?
```
itzg: With the suggested change above, you could wrap these three repetitions in a helper method to stay DRY.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -61,22 +63,38 @@ public String build(String tenantId, String measurement,  TaskParameters taskPar
...
+                String.format("\"info_count\" >= %d", taskParameters.getInfo().getConsecutiveCount())
                 : null)
         .warnCount(
-          taskParameters.getConsecutiveWarning() != null ?
-            String.format("\"state_count\" >= %d",taskParameters.getConsecutiveWarning())
+          taskParameters.getWarning() != null ?
+            String.format("\"warn_count\" >= %d",taskParameters.getWarning().getConsecutiveCount())
             : null)
-        .critCount(String.format("\"state_count\" >= %d", taskParameters.getConsecutiveCritical()))
+        .critCount(String.format("\"state_count\" >= %d", taskParameters.getCritical().getConsecutiveCount()))
```
itzg: If you add an `.as('crit_count')` in the template you can be consistent with the expression structure here...and then probably refactor the three of these with a helper method.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -61,22 +63,38 @@ public String build(String tenantId, String measurement,  TaskParameters taskPar
...
+    levelExpression = taskParameters.getInfo();
     return taskTemplate.execute(TaskContext.builder()
         .labels(taskParameters.getLabelSelector() != null ? taskParameters.getLabelSelector().entrySet() : null)
-        .alertId(taskIdGenerator.generateAlertId(tenantId, measurement, taskParameters.getField()))
+        .alertId(taskIdGenerator.generateAlertId(tenantId, measurement, taskParameters.getCritical().getExpression().getField()))
         .labelsAvailable(labelsAvailable)
         .measurement(measurement)
         .details("task={{.TaskName}}")
-        .critExpression(String.format("\"%s\" %s %s", taskParameters.getField(), taskParameters.getComparator(), taskParameters.getThreshold()))
+        .critExpression(String.format("\"%s\" %s %s",
```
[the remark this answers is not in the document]
harageth: Yeah I was considering that. That brings up a thought that perhaps I should create a validator to make sure that one of the three is set? It just doesnt make sense to create a task that doesnt have any criteria build into it (for a moment I thought about flapping detection but that relies on the criteria too).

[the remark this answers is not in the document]
itzg: Ah yes, that's really what we're wanting to enforce: one or more level are declared. I'd say create another one-off validation, but rest assured we can eventually refactor our validators into a general purpose "one of fields set" and "one or more fields set" validator.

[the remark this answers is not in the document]
harageth: I do also want to note in case it wasn't clear what my intentions were before, that if the user wants the same criteria for each this isn't detecting that and is creating each individually. The intention is that we do a performance run in the future for that sort of detection and simplification to the TICK script.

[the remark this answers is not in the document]
itzg: Ah, I had missed that aspect, so glad you pointed it out. That seems a reasonable strategy. From the REST API response at least, the nodes to provide a `avg_exec_time_ns` metric so you might actually be able to do quick comparison of the "optimized" vs as-is.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -61,25 +64,38 @@ public String build(String tenantId, String measurement,  TaskParameters taskPar
...
+    levelExpression = taskParameters.getInfo();
     return taskTemplate.execute(TaskContext.builder()
         .labels(taskParameters.getLabelSelector() != null ? taskParameters.getLabelSelector().entrySet() : null)
-        .alertId(taskIdGenerator.generateAlertId(tenantId, measurement, taskParameters.getField()))
+        .alertId(taskIdGenerator.generateAlertId(tenantId, measurement, taskParameters.getCritical().getExpression().getField()))
         .labelsAvailable(labelsAvailable)
         .measurement(measurement)
         .details("task={{.TaskName}}")
-        .critExpression(String.format("\"%s\" %s %s", taskParameters.getField(), taskParameters.getComparator(), taskParameters.getThreshold()))
+        .critExpression(taskParameters.getCritical() != null ?
```
itzg: You could take the refactor one level higher, have `buildTICKExpression` take a `LevelExpression`, and call this as

```
.critExpression(buildTICKExpression(taskParameters.getCritical()))
```

then, this ternary can move into `buildTICKExpression`.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
itzg: You got the template change but missed:

```suggestion
        .critCount(String.format("\"crit_count\" >= %d", taskParameters.getCritical().getConsecutiveCount()))
```

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
harageth: Oh duh... I had somehow gotten it in my head that I couldn't put the full thing in there because of something...

src/main/java/com/rackspace/salus/event/manage/model/validator/TaskParametersValidator.java
```diff
@@ -0,0 +1,50 @@
...
+import com.rackspace.salus.event.manage.model.TaskParameters;
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+import javax.validation.Constraint;
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import javax.validation.Payload;
+
+public class TaskParametersValidator implements ConstraintValidator<TaskParametersValidator.AtLeastOneOf, TaskParameters> {
```
itzg: It would be really good to add a unit test for the new validator. I just recently figured out how to craft unit tests for those:

https://github.com/racker/salus-telemetry-model/blob/master/src/test/java/com/rackspace/salus/telemetry/model/ValidLabelKeysValidatorTest.java#L31

harageth:
> While commenting on the validator unit test I noticed these changes are lacking unit tests in general...but is that because this is a branch of the other PR?

Yes. My plan was to alter the unit tests for the TICK Script builder once on the main branch (it should currently fail to compile actually because of the altered data types)

src/main/java/com/rackspace/salus/event/manage/model/Expression.java
```diff
@@ -0,0 +1,16 @@
...
+import javax.validation.constraints.NotEmpty;
+import javax.validation.constraints.NotNull;
+import lombok.Data;
+
+@Data
+public class Expression {
+  @NotEmpty
+  String field;
+  @NotNull
+  Number threshold;
```
jjbuchan: Why are we using `Number` vs. something more specific?

src/main/java/com/rackspace/salus/event/manage/model/validator/TaskParametersValidator.java
```diff
@@ -0,0 +1,50 @@
...
+
+import com.rackspace.salus.event.manage.model.TaskParameters;
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+import javax.validation.Constraint;
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import javax.validation.Payload;
+
```
jjbuchan: Can we get a short description up here describing what this validator does.

src/main/java/com/rackspace/salus/event/manage/model/validator/TaskParametersValidator.java
```diff
@@ -0,0 +1,50 @@
...
+    } else {
+      return true;
+    }
+  }
+
+  @Target({TYPE, ANNOTATION_TYPE}) // class level constraint
+  @Retention(RUNTIME)
+  @Constraint(validatedBy = TaskParametersValidator.class) // validator
+  @Documented
+  public @interface AtLeastOneOf {
```
jjbuchan: I think I like that you've included this in here vs. two separate files like our other validators.

src/main/java/com/rackspace/salus/event/manage/services/TasksService.java
```diff
@@ -66,9 +66,9 @@ public TasksService(EventEnginePicker eventEnginePicker, RestTemplateBuilder res
 
   public EventEngineTask createTask(String tenantId, CreateTask in) {
 
-    if (!Comparator.valid(in.getTaskParameters().getComparator())) {
+    if (!Comparator.valid(in.getTaskParameters().getCritical().getExpression().getComparator())) {
```
jjbuchan: Should this be checking all the task parameters?

jjbuchan: Maybe it should just be another validator on the task params?

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -61,25 +64,35 @@ public String build(String tenantId, String measurement,  TaskParameters taskPar
     boolean labelsAvailable = false;
     if(taskParameters.getLabelSelector() != null)
       labelsAvailable = true;
+    levelExpression = taskParameters.getInfo();
     return taskTemplate.execute(TaskContext.builder()
         .labels(taskParameters.getLabelSelector() != null ? taskParameters.getLabelSelector().entrySet() : null)
-        .alertId(taskIdGenerator.generateAlertId(tenantId, measurement, taskParameters.getField()))
+        .alertId(taskIdGenerator.generateAlertId(tenantId, measurement, taskParameters.getCritical().getExpression().getField()))
```
jjbuchan: What is the `alertId` used for?

It seems safe enough to assume that the ok/warning thresholds will relate to the same field as the critical one.  Am I right in thinking that this is why you're just pulling info from `getCritical`?

harageth: That was the assumption I was making when I first made this change. After talking to Mitch it became apparent that the critical expression might not always be set...

src/main/java/com/rackspace/salus/event/manage/services/TasksService.java
harageth: Doing it as another validator is probably better. It will at least simplify this code here and not bog it down with the same validation three times.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
jjbuchan: Ah yeah, good point.  Earlier today I was thinking you may even just want informational alert for some things.

src/main/java/com/rackspace/salus/event/manage/model/Expression.java
itzg: It's so we can let the JSON deserializing put a float or int depending on what the user provided.

src/main/java/com/rackspace/salus/event/manage/model/validator/TaskParametersValidator.java
itzg: I started that pattern, but I have since decided I didn't like it :). That's why you saw my latest one as two separate classes. The `@<Validator>.<Constraint>` style annotation on the fields started to look verbose to me.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
itzg: The alertId is used by kapacitor to populate the `id` field of the event JSON object, such as:

```json
{
  "id": "aaaaaa:development:0:cpu:usage_user",
  "message": "aaaaaa:development:0:cpu:usage_user is OK",
  "details": "task=aaaaaa-cpu-a31723cd-2dd4-4b38-9fc5-67d70d1ea363",
  "time": "2019-05-22T16:25:40Z",
  "duration": 20000000000,
  "level": "OK",
  "data": {
    "series": [
      {
        "name": "cpu",
        "tags": {
          "agent_discovered_arch": "amd64",
          "agent_discovered_hostname": "MS90HCG8WL",
          "agent_discovered_os": "darwin",
          "agent_environment": "localdev",
          "cpu": "cpu-total",
          "pingable": "false",
          "system_metadata_envoyId": "a22a5f90-7ca9-11e9-903a-6a00027f65d0",
          "system_monitoring_system": "SALUS",
          "system_qualified_account": "UNKNOWN:aaaaaa",
          "system_resource_id": "development:0"
        },
        "columns": [
          "time",
          "usage_guest",
          "usage_guest_nice",
          "usage_idle",
          "usage_iowait",
          "usage_irq",
          "usage_nice",
          "usage_softirq",
          "usage_steal",
          "usage_system",
          "usage_user"
        ],
        "values": [
          [
            "2019-05-22T16:25:40Z",
            0,
            0,
            62.45780722590324,
            0,
            0,
            0,
            0,
            0,
            16.6270783847981,
            20.91511438929866
          ]
        ]
      }
    ]
  },
  "previousLevel": "CRITICAL",
  "recoverable": true
}
```

harageth:
@jjbuchan I added some documentation and the other validator.

@itzg I altered the alertId stuff

src/main/java/com/rackspace/salus/event/manage/model/validator/ExpressionValidator.java
```diff
@@ -0,0 +1,39 @@
...
+ * Validates on object creation whether the validator is one of the four acceptable options that we allow
+ */
+public class ExpressionValidator implements ConstraintValidator<AtLeastOneOf, Expression> {
+
+  @Override
+  public boolean isValid(Expression expression, ConstraintValidatorContext context) {
+    return Comparator.valid(expression.getComparator());
+  }
+
+  @Target({TYPE, ANNOTATION_TYPE}) // class level constraint
```
itzg: ...and you would change the target to `FIELD` for a field-level constraint

PR #10, opened 2019-06-07
Rollback partial kapacitor fails

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-414

# What

If one or more of the discovered kapacitor instances successful accepted a task creation, but one of them failed we were leaving the successful ones configured with the kapacitor task. The entity wouldn't be saved, so there was no way to cleanup the partial configuration.

# How

This change implements a pseudo-transaction-rollback behavior by keeping track of the successful engine instances that got configured and for any subsequent errors invokes the task deletion API of the successful instances. If any of those fail, it just logs the warning and moves on with the others.

## How to test

A unit test was created for the `TasksService` and covers the success and failure cases.


src/main/java/com/rackspace/salus/event/manage/services/TasksService.java
```diff
@@ -117,6 +139,10 @@ public EventEngineTask createTask(String tenantId, CreateTask in) {
     return eventEngineTaskRepository.save(eventEngineTask);
   }
 
+  private void rollbackTask(String kapacitorTaskId, List<EngineInstance> fromInstances) {
+
+  }
+
```
GeorgeJahad: is rollbackTask() needed?

itzg: Oops, no. I started that method before realizing `deleteTaskFromKapacitors` would work.

src/test/java/com/rackspace/salus/event/manage/services/TasksServiceTest.java
```diff
@@ -0,0 +1,280 @@
...
+    verify(eventEnginePicker).pickAll();
+
+    verify(eventEngineTaskRepository).save(eventEngineTaskArgumentCaptor.capture());
+    assertThat(eventEngineTaskArgumentCaptor.getValue().getId())
+        .isEqualTo(taskId.getBaseId());
+
+    verifyNoMoreInteractions(eventEnginePicker, eventEngineTaskRepository, taskIdGenerator,
+        tickScriptBuilder
+    );
+  }
```
GeorgeJahad: should we also be verifying the mockKapacitorServer?:   https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/test/web/client/MockRestServiceServer.html#verify--

itzg: When I had bugs in the code/test it the `expect` calls already seem to be flagging the mismatches...I wonder if Spring Boot's unit test wrapper is calling verify. I'll investigate more.

GeorgeJahad:
Is it worth adding a test for TasksService::deleteTask?

src/test/java/com/rackspace/salus/event/manage/services/TasksServiceTest.java
itzg: ...using the debugger I confirmed the verify wasn't getting called already, so I'm adding these.

itzg:
> Is it worth adding a test for TasksService::deleteTask?

As you can probably, I was adding the bare minimum of unit tests for the new code. But good question, since I'm already in here I'll cover the deleteTask also.

PR #11, opened 2019-06-10
add support for eval and lists of window fields

GeorgeJahad:
# Resolves

https://jira.rax.io/browse/SALUS-367
https://jira.rax.io/browse/SALUS-368

# What

add support for eval and lists of window fields
# How

Added support for evalNodes in tickscripts:
https://docs.influxdata.com/kapacitor/v1.5/nodes/eval_node/

The user presents a list of operands, an operator, and a string to refer to the result "as".
The operand could be a real number, field/tag name, or a
 [built in function](https://docs.influxdata.com/kapacitor/v1.5/tick/expr/#stateful-functions)

The tick script builder normalizes the operands and "as", by quoting appropriately, and validates the function name.

Also added support for lists of "windowFields", from which are generated min, max and avg fields.

## How to test

unit tests
# Why

to allow for more powerful tick scripts to be created by the user.
# TODO
I confirmed that all the current combinations of output nodes are valid by looking at the [mustache file](https://github.com/racker/salus-event-engine-management/blob/george/addEval/src/main/resources/templates/task.mustache) and confirming that    any chained methods following each node is valid.  I still have to look into the difference between the ["where" property method]( https://docs.influxdata.com/kapacitor/v1.5/nodes/from_node#where) and the [where chain method](https://docs.influxdata.com/kapacitor/v1.5/nodes/eval_node/#where)


src/main/java/com/rackspace/salus/event/manage/model/validator/EvalExpressionValidator.java
```diff
@@ -0,0 +1,86 @@
...
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import javax.validation.Payload;
+
+
+public class EvalExpressionValidator implements
+    ConstraintValidator<EvalExpressionValidation, EvalExpression> {
+
+  public static String functionRegex = "(\\w+)\\((.*)\\)";
+  private static List<String> validFunctions = Arrays.asList(
```
itzg: I would have been lazy and just had it confirm it looked like a valid function reference :)

src/main/java/com/rackspace/salus/event/manage/model/validator/EvalExpressionValidator.java
```diff
@@ -0,0 +1,86 @@
...
+      //  and already available in com.rackspace.salus.event.manage.model.Expression
+      //  https://docs.influxdata.com/kapacitor/v1.5/tick/expr/#conditional-functions
+  );
+
+  private boolean isValidOperand(String operand) {
+    if (!Pattern.matches(functionRegex, operand)) {
+      return true;
+    }
+    // confirm function call invokes valid function name
+    return validFunctions.stream().anyMatch(f -> operand.startsWith(f + "("));
```
itzg: Since you have the matching group declared in the regex, you could pick off the function name and then just use [Collection.contains](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#contains-java.lang.Object-). With that line of thought, you could save off the `validFunctions` list into a `HashSet` and then it'll be slightly more efficient.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -91,6 +102,58 @@ public String buildTICKExpression(LevelExpression consecutiveCount, String forma
         null;
   }
 
+  private Boolean isValidRealNumber(String operand) {
+    return Pattern.matches("^[-+]?([0-9]+(\\.[0-9]+)?|\\.[0-9]+)$", operand);
```
itzg: My lazy way to do this is

```
try {
  Double.parseDouble(operand);
  return true;
} catch (NumberFormatExcetion e) {
  return false;
}
```

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -91,6 +102,58 @@ public String buildTICKExpression(LevelExpression consecutiveCount, String forma
...
+    Matcher matcher = Pattern.compile(EvalExpressionValidator.functionRegex).matcher(operand);
+
+    //  if operand is not a function call, double quote it
+    if (!matcher.matches()) {
+      // operand doesn't contain function, and thus is a tag/field name requiring double quotes
+      return "\"" + operand + "\"";
+    }
+
+    // Operand is function call, so split out the function parameters, double quoting the tag/fields
+    String parameters = Arrays.stream(matcher.group(2).split(","))
```
itzg: Nice use of streams

src/test/java/com/rackspace/salus/event/manage/services/TickScriptBuilderTest.java
```diff
@@ -42,8 +45,14 @@
   @Autowired
   TickScriptBuilder tickScriptBuilder;
 
+  private static String readContent(String resource) throws IOException {
```
itzg: Once we get the GCB build going we can switch this over to https://github.com/racker/salus-test/blob/master/src/main/java/com/rackspace/salus/test/JsonTestUtils.java#L47

src/test/java/com/rackspace/salus/event/manage/services/TickScriptBuilderTest.java
```diff
@@ -52,7 +61,7 @@ public void testBuild() throws IOException{
             .setComparator(">")
             .setField("field")
             .setThreshold(33));
-    Map<String, String> labelSelectors = new HashMap();
+    Map<String, String> labelSelectors = new HashMap<>();
```
itzg: Thanks for fixing the warnings.

src/test/resources/TickScriptBuilderTest/testBuildEval.tick
```diff
@@ -0,0 +1,13 @@
+stream
+  |from()
+    .measurement('measurement')
+    .groupBy('resourceId')
+  |eval(lambda: 1.0 + "field" + sigma("cpu", 1), lambda: 2.0 - "tag" - count("field2", 1))
```
itzg: So cool to see it all come together.

GeorgeJahad:
I did a rebase since there were merge conflicts.  here are the changes since then: 
https://github.com/racker/salus-event-engine-management/compare/cd1f5cd...george/addEval

basically just fixing the merge conflicts, updating the readme and changing to use the hashSet

PR #15, opened 2019-06-19
minor bug fix

GeorgeJahad:


GeorgeJahad:
required because of this:
https://github.com/racker/salus-event-engine-management/blob/862b37203e7852b4ca6c089b3922d7476c3a1515/src/test/java/com/rackspace/salus/event/manage/EventEngineManageApplicationTests.java#L27

itzg:
Not that it helps your scenario but:

- GCP builds have been fine [here](https://console.cloud.google.com/cloud-build/builds?query=trigger_id%3D%228faeb2b1-9ce0-4f1b-b519-3ba06683e858%22&project=salus-220516&authuser=1&organizationId=767495606862)
- A local run of `mvn clean install` passed all tests fine

Now that I think about it more, it would be best if not having to add the `spring.application.name` in the test yaml since it could mask a non-test application config issue.

GeorgeJahad:
mvn clean does make it go away, sorry for the confusion.  I'll be more careful about that in the future.

Just fyi,  this is the error I was getting:
java.lang.IllegalStateException: Failed to load ApplicationContext
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'errorAttributes' defined in com.rackspace.salus.common.web.ExtendedErrorAttributesConfig: Unexpected exception during bean creation; nested exception is java.lang.IllegalArgumentException: Could not resolve placeholder 'spring.application.name' in value "${spring.application.name}"

on this test:
https://github.com/racker/salus-event-engine-management/blob/128703bb7d1d406b486651528dd2ba05e868d33f/src/test/java/com/rackspace/salus/event/manage/EventEngineManageApplicationTests.java#L31

GeorgeJahad:
closing, not needed

PR #16, opened 2019-06-24
Add timestamps to events + some cleanup

jjbuchan:
# Resolves

https://jira.rax.io/browse/SALUS-422

# What

Adds create/update timestamps.
Cleans up some old things.

# How

Adds a `toDTO` method to correctly convert from Instant to the timestamp format we want.  Also ensures the right fields are being converted and nothing we care about gets excluded like what may have been happening with the objectmapper.

## How to test

Run tests.
Run locally.

# Why

Everything that can be created/updated should have a timestamp.

### DIFF
--- a/src/main/java/com/rackspace/salus/event/manage/entities/EventEngineTask.java
+++ b/src/main/java/com/rackspace/salus/event/manage/entities/EventEngineTask.java
@@ -16,16 +16,21 @@
 
 package com.rackspace.salus.event.manage.entities;
 
+import com.rackspace.salus.event.manage.model.EventEngineTaskDTO;
 import com.rackspace.salus.event.manage.model.TaskParameters;
 import com.vladmihalcea.hibernate.type.json.JsonStringType;
 
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
 import java.util.Map;
 import java.util.UUID;
 import javax.persistence.*;
 
 import lombok.Data;
+import org.hibernate.annotations.CreationTimestamp;
 import org.hibernate.annotations.Type;
 import org.hibernate.annotations.TypeDef;
+import org.hibernate.annotations.UpdateTimestamp;
 
 @Entity
 @Table(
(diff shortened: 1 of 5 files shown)


src/main/java/com/rackspace/salus/event/manage/model/EventEngineTaskDTO.java
```diff
@@ -17,14 +17,22 @@
...
+import com.rackspace.salus.telemetry.model.View;
 import java.util.UUID;
 import lombok.Data;
 
 @Data
 public class EventEngineTaskDTO {
   UUID id;
+
+  @JsonView(View.Admin.class)
   String tenantId;
```
jjbuchan: In other DTOs I thik we took out this field, but it occurred to me that it'd be good to keep it in for any admin api requests.

We may get back a multi-tenant response so the field would be relevant.  For customers interacting with a single tenant it is of no use.

src/main/java/com/rackspace/salus/event/manage/entities/EventEngineTask.java
```diff
@@ -44,6 +49,9 @@
   @Column(nullable = false)
   String tenantId;
 
+  @Column(nullable = false)
```
itzg: Since we tend not to use this field, we could make it null-able; however, I guess it's not good schema-migration practice to later change that.

jjbuchan: Is this field supposed to be like an ele label?  Just some describing "name" for the task?

I see we also have `monitorName` so assume this is the same.  I'd say making it non-null is a good idea since it being populated helps when it comes to what a notification/ticket can look like.

itzg: Yes, it's supposed to be like the ele label and yes, is the same as the `monitorName` field for monitors.

PR #20, opened 2019-06-26
Rename taskId to kapacitorTaskId

jjbuchan:
Add some clarity to this field and also exclude it from public/internal apis.


jjbuchan:
If we really wanted to rename more and keep things generic I'd probably want to change it to `eventEngineTaskId` and what is currently `EventEngineTask` would just be `Task` or something like that... but I expect we wont bother with all that.

PR #32, opened 2019-12-19
Fail task create/delete when all engines are missing

itzg:
# Resolves

Management part of https://jira.rax.io/browse/SALUS-703

# What

When there was a misconfiguration of kapacitor in staging we realized that it was possible for the API operation to create event tasks to happily succeed even when no kapacitor (aka "engine") instances were discovered.

# How

Checks if the discovered engine list is non-empty before finishing out the create/delete operations.

Also, since we're now using InnoDB in our local MySQL, I added `@Transactional` support to the create/delete service methods. The bulk of the code changes are due to me switching the unit test to using real DB operations rather than mocking the repository interface.

FYI I wrote a new Jira issue https://jira.rax.io/browse/SALUS-728 to document a feature we would need when scaling out kapacitor instances. That feature would actually be another way to solve the case when there's zero kapacitor's and one or more are later added.

## How to test

Added and updated unit tests


src/main/java/com/rackspace/salus/event/manage/services/TasksService.java
```diff
@@ -146,6 +153,7 @@ public EventEngineTask createTask(String tenantId, CreateTask in) {
     return eventEngineTaskRepository.findByTenantId(tenantId, page);
   }
 
+  @Transactional
```
jjbuchan: Should the two delete operations in this method be reversed?

jjbuchan: It might not make a big difference but `deleteTaskFromKapacitors` relies on the object you're deleting in the step beforehand and seems like it would be more likely to fail and trigger the rollback.  Waiting to delete the event task until all kapacitor tasks are removed seems a little better?

src/test/java/com/rackspace/salus/event/manage/services/TasksServiceTest.java
```diff
@@ -250,22 +258,58 @@ private void common_testCreate_fail(ResponseCreator responseCreator) throws IOEx
...
+
+    final CreateTask taskIn = buildCreateTask();
+
+    // EXECUTE
+
+    try {
+      tasksService.createTask("t-1", taskIn);
+
+      fail("Should have thrown IllegalStateException");
+    } catch (IllegalStateException e) { }
```
jjbuchan: I still prefer using exception rules for this type of thing - https://github.com/racker/salus-telemetry-monitor-management/blob/182fba21d0362a0cb87e2e95a01daeb03cfbcf4d/src/test/java/com/rackspace/salus/monitor_management/services/MonitorConversionServiceTest.java#L342

src/main/java/com/rackspace/salus/event/manage/services/TasksService.java
itzg: Like you said, probably doesn't make a difference in the end, but does make more sense since it would match the ordering of the create method. I'll give it a try.

src/test/java/com/rackspace/salus/event/manage/services/TasksServiceTest.java
itzg: Actually I prefer AssertJ's `assertThatThrownBy` which has similar benefits as exceptions rules (but less tedious IMHO). I was tempted to switch these over, but wanted to avoid changing too many lines :) ...that was before I changed all the repository stuff.

PR #39, opened 2020-03-02
handles default consecutive count and handles empty sets of labels

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-824

# What

We wanted to clean up a couple things in the tick script builder code and make sure that there were some cases that were getting handled properly. Specifically we wanted a default consecutive count and needed to make sure that an empty map of labels would be treated correctly.

# How

Makes sure that the swagger template only sets the consecutive count if the expression is created. This also makes sure to detect an empty map of labels and set the values the same as if it were null (which I think is correct). 

## How to test

Run the Unit tests

# Why

The consecutive counts should only be set (and only make sense being set) when their respective expression is created. I could have set a default in the TaskContext object but that would be setting the direct string to be inserted into the tick script. By setting the default on the object coming in from the user we make sure that the value properly gets propagated through everything and goes through all of the validation of everything else.

For handling the empty set of labels we needed to just do the same thing we did in the null case. 

# TODO

Merge in the model PR first


src/main/resources/templates/task.mustache
```diff
@@ -51,15 +51,15 @@ stream
     .stateChangesOnly()
     .id('{{alertId}}')
     .details('{{details}}')
-{{#infoCount}}
+{{#infoExpression}}
     .info(lambda: {{infoCount}})
-{{/infoCount}}
```
itzg: I was expecting this to be:

```
{{#infoCount}}
    .info(lambda: {{infoCount}})
{{/infoCount}}
{{^infoCount}}
    .info(lambda: {{infoExpression}})
{{/infoCount}}
```

and then at https://github.com/racker/salus-event-engine-management/blob/76e7c2b954ef9045ce55f927f0d684e38e5c862b/src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java#L79 it would conditionalize on a null consecutive count as

```
        .infoCount(taskParameters.getInfo().getConsecutiveCount() != null ?
          buildTICKExpression(taskParameters.getInfo(), "\"info_count\" >= %d") : null)
```

While tweaking that last bit of code, I'm wondering if maybe the field could be an `int`, default to zero, and if TICK would skip the `1m` state duration node and fire immediately when the expression evaluates to true.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -102,16 +112,12 @@ public String buildTICKExpression(LevelExpression consecutiveCount, String forma
         null;
   }
 
-  private Boolean isValidRealNumber(String operand) {
-    return Pattern.matches("^[-+]?([0-9]+(\\.[0-9]+)?|\\.[0-9]+)$", operand);
-  }
-
   private String normalize(String operand) {
-    if (isValidRealNumber(operand)) {
+    if (validRealNumber.matcher(operand).matches()) {
```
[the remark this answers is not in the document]
harageth: It just seemed silly to make a function call to effectively just do a `.matches()` (because .matcher(operand) is roughly the equivalent of isValidRealNumber)

I can create the isValidRealNumber function again. I don't feel super strongly about getting rid of it.

src/main/resources/templates/task.mustache
harageth: Alternatively I remember something that zero provided as a count threshold may also skip the duration and just trigger immediately (I'll need to look into that). 

This being said, we should have a default value and it should probably be something between 3  and 5.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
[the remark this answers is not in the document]
itzg: Not that you should bank on it, but the JVM will optimize small, frequently called functions to be inlined anyway. Besides, it's not silly at all to abstract a somewhat unintuitive pattern matches usage by wrapping it in a well named function. Perhaps later we want to swap this with non-regex approach and then only a single spot needs to be changed.

src/main/resources/templates/task.mustache
harageth: https://docs.influxdata.com/kapacitor/v1.5/nodes/state_duration_node/

A value of zero will trigger immediately

itzg: Excellent. Thanks for confirming since that'll make the template conditionals much easier.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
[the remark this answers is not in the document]
GeorgeJahad: I feel like small well named functions are often better than comments, and always consider creating one instead of adding a comment.

src/test/java/com/rackspace/salus/event/manage/services/TickScriptBuilderTest.java
```diff
@@ -179,15 +197,35 @@ public void testEvalExpression() throws IOException {
...
+    String expectedString = readContent("/TickScriptBuilderTest/testNullConsecutiveCount.tick");
+    Map<String, String> labelSelectors = new HashMap<>();
+    labelSelectors.put("resource_metadata_os", "linux");
+    labelSelectors.put("resource_metadata_env", "prod");
+
+    LevelExpression critExpression = new LevelExpression();
+    critExpression.setExpression(new Expression()
+            .setComparator(">")
+            .setField("field")
+            .setThreshold(33));
```
itzg: In case we change our mind again :) , please explicitly set `stateDuration` to null.

itzg:
For traceability, I wanted to note that this PR depends on https://github.com/racker/salus-telemetry-model/pull/117

src/test/java/com/rackspace/salus/event/manage/services/TickScriptBuilderTest.java
harageth: Reasonable

harageth: Actually that won't work. If we set it to null directly, we ignore the default value being set in the object itself. I didn't actually add any logic to here to handle null values of the stateDuration. Everything assumes that with the default value on the object that stateDuration will be set.

itzg: I'm confused now. It looks like the consensus (https://github.com/racker/salus-telemetry-model/pull/117#discussion_r387322775) is to default `stateDuration` to null. As such, null will need to be properly handled by the task generator.

harageth: oh I suppose it would have helped to have seen James comment.

itzg: I was wondering if that was the case :)

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -107,8 +107,8 @@ public String buildTICKExpression(LevelExpression expression) {
         null;
   }
 
-  public String buildTICKExpression(LevelExpression consecutiveCount, String formatString) {
-    return consecutiveCount != null ? String.format(formatString, consecutiveCount.getStateDuration()) :
+  public String buildTICKExpression(LevelExpression stateDuration, String formatString) {
```
itzg: To avoid confusion, maybe this parameter should be called "levelExpression".

src/main/resources/templates/task.mustache
```diff
@@ -32,34 +32,55 @@ stream
     .as({{joinedAs}})
     .keep()
 {{/joinedEvals}}
-{{#critExpression}}
+{{#critCount}}
+  {{#critExpression}}
```
itzg: I'm wondering if just testing for `critCount` would have been sufficient, but evaluating both conditions does seem good anyway.

src/test/java/com/rackspace/salus/event/manage/services/TickScriptBuilderTest.java
```diff
@@ -220,6 +220,7 @@ public void testNullConsecutiveCount() throws IOException {
             .setComparator(">")
             .setField("field")
             .setThreshold(33));
+            //.setStateDuration(null);
```
itzg: Uncomment this, right?

PR #45, opened 2020-05-27
Adam/task id fix

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-884

# What

This PR is making sure that we can create Tasks for "hybrid:xxxxxx" style tenants

# How

It alters how we are generating the taskId so that we don't use restricted characters

## How to test

The only way to test this is to send the completed script to Kapacitor and have it tell us what is and isn't restricted. We don't have that capability in any of our current unit tests.

The best way to test this is to start this app and create a task with the "hybrid:xxxxxx" tenant id. 

# Why

I have attempted to look at Kapacitor to see if there is a better way to alter restricted characters within the task id but haven't been able to find anything. Short of altering the kapacitor config itself the only thing we can do is alter our id generation methods.

### DIFF
--- a/src/main/java/com/rackspace/salus/event/manage/services/KapacitorTaskIdGenerator.java
+++ b/src/main/java/com/rackspace/salus/event/manage/services/KapacitorTaskIdGenerator.java
@@ -22,6 +22,8 @@ import org.springframework.stereotype.Component;
 
 @Component
 public class KapacitorTaskIdGenerator {
+  private static final String UNDERSCORE = "_";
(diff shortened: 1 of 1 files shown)


harageth:
> Assuming we don't try to reuse this value later (which I don't think we should) this works for me.
> 
> A little torn between stripping the `:` vs. replacing it but in the end it shouldn't matter.

I am not super happy with it either. I would have preferred being able to just alter the config for kapacitor later. We could also change the UNDERSCORE value to something more generic and load it in from a config.

PR #48, opened 2020-06-23
Added test-event-task feature

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-936

# What

Provides an endpoint and service for performing tests of new event-tasks given a task definition and a set of metrics obtained from a test-monitor result.

The associated proposal is located here, where the sequence diagram is most beneficial to follow along when reviewing:

https://github.com/Rackspace-Segment-Support/salus-docs-internal/tree/master/proposals/test-event-task

## How to test

Unit tests added and was also tested with an associated Insomnia workspace addition.

# Depends on

- https://github.com/racker/salus-common/pull/68


src/main/java/com/rackspace/salus/event/manage/web/controller/TasksApiController.java
```diff
@@ -92,4 +100,28 @@ public void deleteTask(@PathVariable String tenantId,
...
+    return testEventTaskService.performTestTask(tenantId, request)
+        .thenApply(testTaskResult ->
+            ResponseEntity
+                .status(deriveTestTaskStatus(testTaskResult))
+                .body(testTaskResult));
+  }
+
+  private HttpStatus deriveTestTaskStatus(TestTaskResult testTaskResult) {
+    if (StringUtils.hasText(testTaskResult.getError())) {
+      return HttpStatus.UNPROCESSABLE_ENTITY;
```
jjbuchan, line 122: I wonder if 504 is more likely here?  We'd probably have to inspect the error details to know exactly what went wrong.

Knowing when timeouts happen vs. other errors could be helpful so it might be worth looking into.  But other than that reason I could live with the 422.

src/main/java/com/rackspace/salus/event/manage/model/ValidationGroup.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rackspace.salus.event.manage.model;
+
+public class ValidationGroup {
```
jjbuchan, line 19: nit: we call this `ValidationGroups` in other repos

src/test/java/com/rackspace/salus/event/manage/services/TestEventTaskServiceTest.java
```diff
@@ -0,0 +1,454 @@
...
+    final CompletableFuture<TestTaskResult> completableResult = testEventTaskService
+        .performTestTask("t-1", request);
+
+    // ...and do not send kapacitor event via kafka
+
+    // Assertions
+
+    final TestTaskResult result = completableResult.get(5, TimeUnit.SECONDS);
+    assertThat(result).isNotNull();
+    assertThat(result.getError()).isEqualTo("Timed out waiting for test-event-task result");
```
jjbuchan, line 236: This is the kind of thing that I think is possibly a 504 in the api response?

src/main/java/com/rackspace/salus/event/manage/services/TestEventTaskService.java
```diff
@@ -0,0 +1,421 @@
...
+
+    final Point point = pointBuilder.build();
+
+    log.debug(
+        "Posting metrics to instance={} for test with id={}: {}", engineInstance, taskId, point);
+
+    final ResponseEntity<String> response;
+    try {
+      final RequestEntity<String> requestEntity = RequestEntity.post(
+          uriFactory.expand(
```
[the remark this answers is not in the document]
itzg: Really I just wanted to set the content-type header to non-json and then accidentally went down this path of `RequestEntity`. I'll switch this over to the `HttpEntity` and `UriComponentsBuilder.fromUriString` approach for consistency...since I don't know which approach is "better".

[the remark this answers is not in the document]
itzg: ...actually I'm going to do a mixture: the RequestEntity is nicer with the content-type header abstraction, but I'm switching to `UriComponentsBuilder.fromUriString` so I can get rid of the `DefaultUriBuilderFactory` field.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -67,7 +70,17 @@ public TickScriptBuilder(KapacitorTaskIdGenerator kapacitorTaskIdGenerator,
     }
   }
 
-  public String build(String tenantId, String measurement, EventEngineTaskParameters taskParameters) {
+  public String build(String tenantId, String measurement,
+                      EventEngineTaskParameters taskParameters) {
+    return build(tenantId, measurement, taskParameters, "events",
```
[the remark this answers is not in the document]
itzg: That's a good idea; however, maybe I should make it a property since it needs to match the deployment config here

https://github.com/Rackspace-Segment-Support/helm-salus-kapacitor/blob/master/config_files/kafka-events.yml#L2

That would match what I did with the test-task topic:
https://github.com/racker/salus-event-engine-management/blob/7ee2f70a374a622a282051a7784280e80493ad28/src/main/java/com/rackspace/salus/event/manage/config/TestEventTaskProperties.java#L43

src/main/java/com/rackspace/salus/event/manage/web/controller/TasksApiController.java
itzg: I know, I wasn't totally sure what the correct status code should be. 504 does make sense, but the 422 was basically just to be consistent with test-monitor https://github.com/racker/salus-telemetry-monitor-management/pull/71

At least the error's message itself will help give troubleshooting hints regardless of status code.

src/main/java/com/rackspace/salus/event/manage/model/ValidationGroup.java
itzg: Thanks, I'll fix that

src/test/java/com/rackspace/salus/event/manage/services/TestEventTaskServiceTest.java
itzg: Oh yeah, now that I see this context too, a 504 would make sense. Now I'm pondering how to differentiate in the controller...I could add a "timedOut" field to the `TestTaskResult` and exclude it from JSON encoding.

jjbuchan: Is this not the case where a `RestExceptionHandler` would take care of the different response codes as long as we define distinct exceptions for the failures in the service layer?

itzg: Good question. The async/CompletableFuture may or may not change that behavior, but that's why I originally erred on the side of using ResponseEntity explicitly.

itzg: The async exception handling was able to translate the exceptions into status codes as we'd expect.

src/main/java/com/rackspace/salus/event/manage/services/BackendException.java
```diff
@@ -16,8 +16,11 @@
 
 package com.rackspace.salus.event.manage.services;
 
+import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.ResponseStatus;
 
+@ResponseStatus(HttpStatus.BAD_GATEWAY)
```
jjbuchan, line 23: Hmm, I forgot you could add it here without needing to set it on an `AbstractRestExceptionHandler`.  Do you remember why we don't do that in more places?  We have a bunch of things in various `RestExceptionHandler`s.

src/main/java/com/rackspace/salus/event/manage/services/BackendException.java
```diff
@@ -16,8 +16,11 @@
 
 package com.rackspace.salus.event.manage.services;
```
jjbuchan, line 17: Should this be in a different package?  `errors` or `models`?

jjbuchan: Maybe even under `web`?

src/main/java/com/rackspace/salus/event/manage/services/BackendException.java
itzg: Most or all are exception classes that are not owned by us :(

src/main/java/com/rackspace/salus/event/manage/services/BackendException.java
itzg: I'd prefer a new package, `errors`, since a package dependency from `services` -> `web` is "backwards".

src/main/java/com/rackspace/salus/event/manage/services/BackendException.java
itzg: ...I wonder though if we miss out on the enrichment, such as traceId

https://github.com/racker/salus-common/blob/2317222f02ce4d51dbb188fafaada9ca34fe8d05/src/main/java/com/rackspace/salus/common/web/AbstractRestExceptionHandler.java#L99-L102

jjbuchan: > Most or all are exception classes that are not owned by us :(

I was looking here - https://github.com/racker/salus-event-engine-management/blob/master/src/main/java/com/rackspace/salus/event/manage/web/controller/RestExceptionHandler.java#L39 - which is our own, and there's a few others.

If we do miss out on traceId that's definiteyl good reason to move things over.  Otherwise we could probably clean up a lot of older stuff by changing it to what you've done here.

itzg: Yeah, we lose traceId enrichment. I'll switch these over to be exception handler methods instead of the annotation.

PR #49, opened 2020-06-30
Support logical operations in task criteria

jjbuchan:
# What

Updates task expressions to support logical operations.

Utilizes the changes in https://github.com/racker/salus-telemetry-model/pull/139

# How

See `get_task_response.json` for an example of what the payloads will now look like.  When specifying different expressions you need to specify it's type (`comparison` or `logical`) similar to how the `type` value provided in new monitor requests relates to the specific subclass representing each plugin.

Various validators were able to be removed since the restrictions are now better defined by the class structure itself.

# Why

Allows us to migrate over all ele alarms, such as:

```
if (code != 200) {
  return CRITICAL;
}
if (body regex "error" != "") {
  return CRITICAL;
}
return OK;
```

which can now be combined to effectively be represented like

```
if (code != 200 || body regex "error" != "") return CRITICAL
```

This is needed since TickScript can only have one statement per task state.  This expands on ele's functionality by allowing nested logical expressions like

```
if ((code != 200 || code !=500) && body == "") {
  return CRITICAL;
}
```


src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -116,30 +126,66 @@ public String build(String tenantId,
...
+  }
+
+  private String buildTICKExpression(LogicalExpression expression) {
+    String logicalOperator = expression.getOperator().toString();
 
-    return tickExpression;
+    StringBuilder tickExpression = new StringBuilder("(");
+    ListIterator<Expression> iterator = expression.getExpressions().listIterator();
+    while (iterator.hasNext()) {
+      if (iterator.hasPrevious()) {
```
itzg, line 158: Cool, I didn't know about ListIterator and this feature.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -116,30 +126,66 @@ public String build(String tenantId,
...
     }
+  }
+
+  private String buildTICKExpression(LogicalExpression expression) {
+    String logicalOperator = expression.getOperator().toString();
 
-    return tickExpression;
+    StringBuilder tickExpression = new StringBuilder("(");
+    ListIterator<Expression> iterator = expression.getExpressions().listIterator();
+    while (iterator.hasNext()) {
```
itzg, line 157: Since you're a streams guru now, would it work to use `Collectors.joining` to combine these expressions into a string?

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -116,30 +126,66 @@ public String build(String tenantId,
...
+    StringBuilder tickExpression = new StringBuilder("(");
+    tickExpression.append("\"").append(expression.getMetricName()).append("\"");
+    tickExpression.append(" ");
+    tickExpression.append(expression.getComparator());
+    tickExpression.append(" ");
+
+    if (expression.getComparisonValue() instanceof Number) {
+      tickExpression.append(expression.getComparisonValue());
+    } else if (expression.getComparisonValue() instanceof String) {
+      // TODO: booleans will fall under string logic?
```
itzg, line 179: They seem to have a boolean type FWIW

https://docs.influxdata.com/kapacitor/v1.5/tick/syntax/#booleans

src/test/java/com/rackspace/salus/event/manage/services/TestEventTaskServiceTest.java
```diff
@@ -342,7 +343,7 @@ private void verifyTickScriptBuilder(TestTaskRequest request) {
   }
 
   private Task setupMockKapacitorServer(String expectedTaskCreate, String taskId,
-                                        boolean includeTaskGet, boolean failedMetricWrite)
+      boolean includeTaskGet, boolean failedMetricWrite)
```
itzg, line 346: We must have different auto-formatting configured in our IDEs. Either indentation works for me.

src/test/java/com/rackspace/salus/event/manage/services/TickScriptBuilderTest.java
```diff
@@ -110,51 +126,65 @@ public void testBuildOnlyInfo() throws IOException {
...
+            .setState(TaskState.CRITICAL)
+            .setMessage("critical threshold hit"),
+        new StateExpression()
+            .setExpression(warnExpression)
+            .setState(TaskState.WARNING)
+            .setMessage("warning threshold hit"),
+        new StateExpression()
+            .setExpression(infoExpression)
+            .setState(TaskState.INFO)
+            .setMessage("thresholds returned to normal"));
```
itzg, line 156: Can you add a second one of the same state to test against my`.append(" || ")` comment?

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -116,30 +126,66 @@ public String build(String tenantId,
...
+    if (expressions == null || expressions.isEmpty()) {
       return null;
     }
 
-    String tickExpression = String.format("\"%s\" %s ", expression.getExpression().getField(),
-        expression.getExpression().getComparator());
+    StringBuilder tickExpression = new StringBuilder();
+    for (StateExpression stateExpression : expressions) {
+      tickExpression.append(buildTICKExpression(stateExpression.getExpression()));
+    }
```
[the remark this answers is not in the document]
jjbuchan: It's handled by `.append(logicalOperator)` since they use `OR` and `AND` - https://docs.influxdata.com/kapacitor/v1.5/tick/syntax/#operators

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
jjbuchan: Never used it before either.  Glad I found it.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
jjbuchan: Yep, it can.  Now my list iterator finding is worthless!

Looks a little cleaner now, thanks.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
jjbuchan: I'd forgot about this note until pushing the PR.  I might have to rethink this or at least test the logic a little more.

e.g. does a payload of `true` pass both a boolean and string `instanceof` test?  If it does I'm not sure if we can make assumptions about what is expected.

src/test/java/com/rackspace/salus/event/manage/services/TickScriptBuilderTest.java
jjbuchan: Good catch.  Thought I had that already.

Making the change did make me realize `message` probably needs to move up a level, similar to how i moved the different `stateDuration` fields up a level.  Right now it's ignored so doesn't affect anything.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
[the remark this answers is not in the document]
jjbuchan: I see what you mean now.  How it's configured now should work for any brand new tasks created but doesn't ease the creation of things during the migration.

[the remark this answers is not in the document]
jjbuchan: All good now.

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
itzg: > Now my list iterator finding is worthless!

At least you had your moment of glory ;)

src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
itzg: As long as it's not quoted in the API call's payload, Jackson Json should deserialize into a `Boolean`...I think.

jjbuchan: Thankfully, it was that simple.

PR #50, opened 2020-07-02
Redesign eval expressions & handle more custom functions

jjbuchan:
# What

Redesigns how the EvalNodes in the TickScripts are generated and adds the ability to use DerivativeNode

Follows on from https://github.com/racker/salus-event-engine-management/pull/49
Relates to https://github.com/racker/salus-telemetry-model/pull/141

# How

View the `How will functions be handled?` section of https://github.com/Rackspace-Segment-Support/salus-docs-internal/pull/47/files

> Create an abstract class named `CustomMetricExpression` with subclass `EvalNode` and `DerivativeNode`. `EvalNode` will have a subclass `BasicEvalNode`, which acts similar to what we have right now, in addition to `PercentageEvalNode` and any others we decide to create helper classes for.  Each of these classes will have methods that can be called to help generate the TickScript contents.  We will be able to group `EvalNode` expressions together so they can be represented in a format like shown here:
>
> ```go
> stream
>     |eval(lambda: "value" * "value", lambda: 1.0 / "value2")
>         .as('value2', 'inv_value2')
> ```

I moved the `where` section of the tick script up above the `window` nodes.  I have not tested all of these in kapacitor itself, but I feel like that should be the correct place for it.

View `get_task_response.json` for an example of the new payload and `testBuildComplexExpression.tick` for an example of how a tick script may look.

# Why

* Allows us to migrate alarms from ele that use the `percentage` and `rate` functions
* Makes the api request payload more user friendly
* Allows us to easily add more custom functions as needed


itzg:
I think I'll wait to review this one until getting your thoughts on https://github.com/racker/salus-telemetry-model/pull/141#pullrequestreview-443193461

src/main/java/com/rackspace/salus/event/manage/model/ValidationGroups.java
```diff
@@ -16,7 +16,9 @@
 
 package com.rackspace.salus.event.manage.model;
 
+import javax.validation.groups.Default;
```
itzg, line 19: Good to know

PR #52, opened 2020-07-16
Use monitor and zone fields to distinguish between tasks

jjbuchan:
# What

Add monitoring zone and monitor id to all tasks groupBy functions.

## How to test

Updated existing tests and I also tried this with [kapacitor-unit](https://github.com/gpestana/kapacitor-unit) and it seems to work fine for agent measurements that wont have a zone set.

# Why

Each bound monitor per resource should be evaluated individually.  The results for each monitor:resource will then be combined to determine if it should generate a "notification".

# TODO

Probably create a kafka streams app to do the grouping/notify things.

I played around with using `combine`, `cumulativeSum`, `flatten` etc. in the tick script but nothing seems to fully solve our use-case, which is why a separate app will be needed.

### DIFF
--- a/src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
+++ b/src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
@@ -250,7 +250,7 @@ public class TickScriptBuilder {
     @Default
     String details = "";
     @Default
-    String groupBy = Tags.RESOURCE_ID;
+    List<String> groupBy = List.of(Tags.RESOURCE_ID, Tags.MONITOR_ID, Tags.MONITORING_ZONE);
     String joinedEvals;
     String joinedAs;
     DerivativeNode derivative;
--- a/src/main/resources/templates/task.mustache
+++ b/src/main/resources/templates/task.mustache
@@ -1,7 +1,7 @@
 stream
   |from()
     .measurement('{{measurement}}')
-    .groupBy('{{groupBy}}')
+    .groupBy({{#groupBy}}'{{this}}'{{^-last}}, {{/-last}}{{/groupBy}})
 {{#labelsAvailable}}
   |where(lambda:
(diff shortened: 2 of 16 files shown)


src/main/java/com/rackspace/salus/event/manage/services/TickScriptBuilder.java
```diff
@@ -250,7 +250,7 @@ private DerivativeNode getDerivative(List<MetricExpressionBase> customMetrics) {
     @Default
     String details = "";
     @Default
-    String groupBy = Tags.RESOURCE_ID;
+    List<String> groupBy = List.of(Tags.RESOURCE_ID, Tags.MONITOR_ID, Tags.MONITORING_ZONE);
```
itzg, line 253: Eventually we could make this be a task create parameter in the API call, but I agree with keeping this simple and Salus oriented for now. The reason I'm thinking it could be an API parameter is for other monitoring systems where the similar type of grouping keys would have different label/tags in the metrics.

itzg: ...better yet, @OmniBrews , perhaps "monitor ID" and "monitoring zone" should be first-class, top-level fields of the ExternalMetric structure? That also reminds me that we were thinking about cleaning up some field names fairly soon.

jjbuchan: Making a note that all of these fields will be stored in metadata in the ExternalMetric payload.  Neither monitorId or monitoringZone will be top-level fields.

If/when we start handling different monitoring systems we will need a different `groupBy` for each system.

itzg: Aha, this is a Salus-specific default, which is what you suspected in the "lexicon" discussion.

PR #53, opened 2020-07-20
Changed test-task to accept multiple metrics and respond with multiple events

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-947

# What

We later realized that test-task should accept one or more metrics to have feature parity with v1, but also that it will help with some state transition testing.

# How

Adding a metrics list to the test-task request was easy...but it opened up the can of worms where multiple resulting events need to be collected together and only the overall request be completed when all arrive or a timeout is hit.

## How to test

Unit tests have been updated


src/main/java/com/rackspace/salus/event/manage/web/controller/TasksApiController.java
```diff
@@ -117,9 +118,14 @@ public void deleteTask(@PathVariable String tenantId,
...
   })
-  public CompletableFuture<TestTaskResult> testEventTask(
+  public CompletableFuture<ResponseEntity<TestTaskResult>> testEventTask(
       @PathVariable String tenantId,
       @RequestBody @Validated(ValidationGroups.Test.class) TestTaskRequest request) {
-    return testEventTaskService.performTestTask(tenantId, request);
+    return testEventTaskService.performTestTask(tenantId, request)
+        .thenApply(testTaskResult ->
+            testTaskResult.isPartialResults() ?
+                ResponseEntity.status(HttpStatus.PARTIAL_CONTENT).body(testTaskResult) :
```
jjbuchan, line 127: I don't think a 206 is the right code here.  207 would be nice but it doesn't look like we should use that since we're not also in control of the clients.  

I'm between a 200 which would include the partial info and something in the body that shows parts are missing or a 504 to show it didn't entirely complete.

jjbuchan: I see you already have a `partialResults` field.  Changing to a 200 and leaving everything as is could work.

itzg: Yeah, I was pondering simply a 200 there since the body had the partialResults indicator. I'll change to that.

src/main/java/com/rackspace/salus/event/manage/model/TestTaskResult.java
```diff
@@ -26,7 +26,15 @@
...
 @JsonInclude(Include.NON_NULL)
 public class TestTaskResult {
+
+  /**
+   * When true, indicates that the overall operation timed out and <code>events</code>
+   * contains a subset of the expected results.
+   */
   boolean partialResults;
+  /**
+   * Contains an event per metric provided, where .
```
jjbuchan, line 36: where....

itzg: That'll be revealed in the next episode ;)

PR #55, opened 2020-07-27
added getEventTask functionality

shabd67:
# Resolves

[SALUS-569](https://jira.rax.io/browse/SALUS-569)

# What

adding getEventTask functionality to chain test-monitor and test-task operations via single API


src/main/java/com/rackspace/salus/event/manage/web/client/EventTaskApi.java
```diff
@@ -0,0 +1,10 @@
+package com.rackspace.salus.event.manage.web.client;
```
jjbuchan, line 1: Another IntelliJ feature you can utilize is documented [here](https://github.com/Rackspace-Segment-Support/salus-docs-internal/issues/9).  It'll insert our copyright notice into all new files automatically.

src/main/java/com/rackspace/salus/event/manage/web/client/EventTaskApi.java
```diff
@@ -0,0 +1,10 @@
...
+
+import com.rackspace.salus.event.manage.model.TestTaskRequest;
+import com.rackspace.salus.event.manage.model.TestTaskResult;
+
+import java.util.concurrent.CompletableFuture;
+
+public interface EventTaskApi {
+    public CompletableFuture<?> getEventTask(String tenantId, TestTaskRequest request);
```
jjbuchan, line 9: Similar to the monitor PR, I think naming this `performTestTask` might work better

src/main/java/com/rackspace/salus/event/manage/web/client/EventTaskApi.java
jjbuchan: Since these files are now created you might need to manually add the text to each of the new files.

src/main/java/com/rackspace/salus/event/manage/web/client/EventTaskApiClient.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public CompletableFuture<?> getEventTask(String tenantId, TestTaskRequest request) {
+        String uriString = UriComponentsBuilder
+                .fromUriString("/api/tenant/{tenantId}/test-task")
+                .buildAndExpand(tenantId)
+                .toUriString();
+
+        HttpHeaders reqHeaders = new HttpHeaders();
+        reqHeaders.setContentType(MediaType.APPLICATION_JSON);
```
jjbuchan, line 31: Similar to the monitor PR again, including optional additional headers and renaming to `performTestTask` would be good.

src/main/java/com/rackspace/salus/event/manage/web/client/EventTaskApi.java
itzg: With the copyright profile configured, can right click any file and update this way:

image

During a regular commit, can also have IntelliJ run the copyright updates:

image

src/main/java/com/rackspace/salus/event/manage/web/client/EventTaskApiClient.java
jjbuchan: ...as I mentioned in the other PR, we should probably just remove these headers from here.

src/main/java/com/rackspace/salus/event/manage/web/client/EventTaskApi.java
```diff
@@ -1,10 +1,25 @@
+/*
+ * Copyright 2019 Rackspace US, Inc.
```
itzg, line 2: Hmm, the copyright template uses a placeholder `$today.year` to replace the current year here. Can you double check you grabbed the template exactly as provided in https://github.com/Rackspace-Segment-Support/salus-docs-internal/issues/9

src/test/java/com/rackspace/salus/event/manage/web/client/EventTaskApiClientTest.java
```diff
@@ -0,0 +1,90 @@
...
+    String url = String.format("/api/tenant/%s/test-task", tenantId);
+    TestTaskResult testTaskResult = podamFactory.manufacturePojo(TestTaskResult.class);
+    objectMapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
+    mockServer.expect(requestTo(url))
+        .andExpect(method(HttpMethod.POST))
+        .andExpect(content().json(expectedReqJson, true))
+        .andRespond(
+            withSuccess(objectMapper.writeValueAsString(testTaskResult),
+                MediaType.APPLICATION_JSON));
+  }
```
itzg, line 89: Same as https://github.com/racker/salus-telemetry-monitor-management/pull/186#discussion_r463632177 , you're missing the actual call to the client to exercise the unit test.

src/test/java/com/rackspace/salus/event/manage/web/client/EventTaskApiClientTest.java
```diff
@@ -0,0 +1,90 @@
...
+  @MockBean
+  EventEnginePicker eventEnginePicker;
+
+  @Test
+  public void testPerformTestTask() throws IOException {
+    String expectedReqJson = readContent("EventTaskApiClientTest/testPerformTestTask_req.json");
+    String tenantId = RandomStringUtils.randomAlphabetic(8);
+    String url = String.format("/api/tenant/%s/test-task", tenantId);
+    TestTaskResult testTaskResult = podamFactory.manufacturePojo(TestTaskResult.class);
+    objectMapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
```
jjbuchan, line 82: Was there a specific reason you needed to add this?

src/test/java/com/rackspace/salus/event/manage/web/client/EventTaskApiClientTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+package com.rackspace.salus.event.manage.web.client;
+
+import static com.rackspace.salus.common.util.SpringResourceUtils.readContent;
+import static org.springframework.test.web.client.match.MockRestRequestMatchers.content;
+import static org.springframework.test.web.client.match.MockRestRequestMatchers.method;
+import static org.springframework.test.web.client.match.MockRestRequestMatchers.requestTo;
+import static org.springframework.test.web.client.response.MockRestResponseCreators.withSuccess;
+
+import com.fasterxml.jackson.databind.*;
```
jjbuchan, line 25: I think you need to do the Optimize Imports for this PR too.

src/test/java/com/rackspace/salus/event/manage/web/client/EventTaskApiClientTest.java
shabd67: its complaining

"No serializer found for class java.lang.Object and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationFeature.FAIL_ON_EMPTY_BEANS) (through reference chain: com.rackspace.salus.event.manage.model.TestTaskResult["events"]->java.util.ArrayList[0]->com.rackspace.salus.event.manage.model.TestTaskResult$EventResult["data"]->com.rackspace.salus.event.manage.model.kapacitor.KapacitorEvent$EventData["series"]->java.util.ArrayList[0]->com.rackspace.salus.event.manage.model.kapacitor.KapacitorEvent$SeriesItem["values"]->java.util.ArrayList[0])"

jjbuchan: Ok, so that is saying it fails to serialize the field `private List<Object> values;` which is part of `SeriesItem`.

It's odd that having this `configure` below the `podam` request makes it work because I'd assume it is the podam request that is triggering the failure?

Podam has the ability to be given a "helper" method to tell it how to populate `Object` fields.  You can see an example of the method here - https://github.com/racker/salus-telemetry-model/blob/master/src/main/java/com/rackspace/salus/telemetry/entities/EventEngineTaskParameters.java#L126-L134

and an example of how to use it in the test here - https://github.com/racker/salus-event-engine-management/blob/master/src/test/java/com/rackspace/salus/event/manage/web/model/EventEngineTaskDTOTest.java#L38-L46

Once you add those items to this case I'd assume you might not need that option.

jjbuchan: @itzg do you agree with that or do you think its worth disabling the `FAIL_ON_EMPTY_BEANS`?

src/main/java/com/rackspace/salus/event/manage/web/client/EventTaskApi.java
```diff
@@ -1,8 +1,8 @@
 /*
- * Copyright 2019 Rackspace US, Inc.
+ * Copyright 2020 Rackspace US, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
+ * you may not use this filee except in compliance with the License.
```
itzg, line 5: Not a big deal to fix this; however, check that your template doesn't have this typo.

src/test/java/com/rackspace/salus/event/manage/web/client/EventTaskApiClientTest.java
itzg: I agree it's best to avoid touching serialization options. This might be another case where it's overly complicated to get podam generated POJOs to cooperate and just hardcode the POJO content instead. Changing to @jjbuchan 's suggestion would be good practice though.

jjbuchan: Yeah, I almost suggested that for a different test.  @shabd67 You'll see some other tests just create the object like so

```
final TestTaskResult expectedResult = new TestTaskResult()
        .setEvents(List.of(
            new EventResult()
            .setData(
                new EventData()
                .setSeries(List.of(
                    new SeriesItem()
                        .setName(testTaskRequest.getTask().getMeasurement())
                ))
            )
            .setLevel("CRITICAL")
        ))
        .setStats(
            new Stats()
            .setNodeStats(Map.of("alert2", Map.of("crits_triggered", 1)))
        );
```

PR #60, opened 2020-08-07
structured partial responses into errors array

shabd67:
# Resolves

salus [989](https://jira.rax.io/projects/SALUS/issues/SALUS-989)

# What

structure partial responses into errors array


src/main/java/com/rackspace/salus/event/manage/model/TestTaskResult.java
```diff
@@ -24,26 +24,30 @@
...
-   */
-  boolean partialResults;
-  /**
-   * Contains the resulting event per metric provided in the request.
-   */
-  List<EventResult> events;
-  /**
-   * Contains the kapacitor stats for task that was created.
-   */
-  Task.Stats stats;
```
itzg, line 42: Same kind of mass-whitespace change as https://github.com/racker/salus-telemetry-monitor-management/pull/190#discussion_r467133215

PR #61, opened 2020-08-12
Add alert message to kapacitor task

shabd67:
# Resolves

[SALUS-952](https://jira.rax.io/browse/SALUS-952)

# What

Add alert message to kapacitor task


src/main/resources/templates/task.mustache
```diff
@@ -66,6 +66,9 @@ stream
     .stateChangesOnly()
 {{/stateChangesOnly}}
     .id('{{alertId}}')
+    {{#messageText}}
```
itzg, line 69: Ah, is this what resolve the "No key, method or field with name 'messageText'" which I'm guessing happened when `messageText` was null?

shabd67: it was due to simplifyTask(). adding messageTemplate to taskParameters solved the problem.

src/test/resources/TickScriptBuilderTest/testAddAlertMessageToKapacitor.tick
```diff
@@ -0,0 +1,18 @@
...
+  |where(lambda:
+    isPresent("resource_metadata_os") AND ("resource_metadata_os" == 'linux')
+      )
+  |stateDuration(lambda: ("field" > 0))
+    .unit(1m)
+    .as('crit_count')
+  |alert()
+    .stateChangesOnly()
+    .id('{{ .TaskName }}:{{ .Group }}')
+      .message('The CPU usage was too high')
```
harageth, line 14: nit: Can this be at the same tabulation as whats around it?

src/main/resources/templates/task.mustache
```diff
@@ -66,6 +66,9 @@ stream
...
+      .message('{{messageText}}')
+    {{/messageText}}
```
itzg, line 71: Nit: can you shift these lines left so that `.message` lines up with `.id` and `.stateChangesOnly`?

src/test/resources/TickScriptBuilderTest/testAddAlertMessageToKapacitor.tick
itzg: Phew, I wasn't the only one picking on that https://github.com/racker/salus-event-engine-management/pull/61#discussion_r469405722

harageth: I like that we even used the same `nit` notation

itzg:
PR triggered builds are still failing, but investigating via https://github.com/racker/salus-event-engine-management/pull/63 has shown inconsistent behavior of cloud build. So I am going to merge this and we can research more at a later time.

jjbuchan:
I thought Kapacitor had the option to add a different message per state but I'm not seeing that in the docs now, so this is good!

itzg:
> I thought Kapacitor had the option to add a different message per state but I'm not seeing that in the docs now, so this is good!

Now that you mentioned it, I thought I had too, but like you I just now looked and only see the `critReset` type expression that we haven't used yet.

PR #62, opened 2020-08-12
Adam/delete tenant

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-956

# What

Delete all tasks for a particular tenant

# How

Utilizes the new repository function to delete all tasks associated with a tenant both in the database and in all kapacitor instances.

## How to test

Unit tests have been added

# Why

It seems silly to not do this and instead delete everything individually.


src/main/java/com/rackspace/salus/event/manage/services/TasksService.java
```diff
@@ -196,4 +197,15 @@ private void deleteTaskFromKapacitors(String kapacitorTaskId,
...
   }
+
+  @Transactional
+  public void deleteAllTasksForTenant(String tenant) {
+    Page<EventEngineTask> results = eventEngineTaskRepository
+        .findByTenantId(tenant, Pageable.unpaged());
+    Collection<EngineInstance> eventEngines = eventEnginePicker.pickAll();
+    results.forEach(value -> deleteTaskFromKapacitors(value.getKapacitorTaskId(),
+        eventEngines, false));
+
```
GeorgeJahad, line 208: if this fails in the middle, then some kapacitor tasks would be deleted and some not.

if the user then retried the request, would it then fail because the ones that were deleted, but are still in the db, cause it to continue to fail?

In other words, would it be safer to delete each db entry as each kapacitor task is deleted?

harageth: Yes you are correct. And we should start a transaction for each deletion and only submit the transaction when the kapacitor task has been deleted too.

itzg:
@harageth there's more incoming changes on this one, right?

harageth:
> @harageth there's more incoming changes on this one, right?

It is updated now

PR #65, opened 2020-08-21
Add the ability to update/PUT an event task

shabd67:
# Resolves

[SALUS-997](https://jira.rax.io/browse/SALUS-997)

# What

Add the ability to update/PUT an event task

# Depends on

https://github.com/racker/salus-telemetry-api/pull/100


src/test/java/com/rackspace/salus/event/manage/services/TasksServiceTest.java
```diff
@@ -103,9 +107,18 @@
...
+
+  @Autowired
+  EntityManager entityManager;
+  @Autowired
+  JdbcTemplate jdbcTemplate;
+
   @After
   public void tearDown() throws Exception {
     eventEngineTaskRepository.deleteAll();
+    entityManager.flush();
```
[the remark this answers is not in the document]
jjbuchan: @shabd67 for future reference, these types of comments aren't necessarily asking you to remove/change the code.  We've hit some weird issues in the past that has led to things like this being added and in some cases we might not always fully understand why its needed.

If there's wasn't a specific need it's fine to push a commit to remove it, but following up with a response to say yes/ no and any reasoning is appreciated too.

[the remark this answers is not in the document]
jjbuchan: i.e. PR reviews can sometimes include general dialog vs. requests/suggestions for change.

src/main/java/com/rackspace/salus/event/manage/web/controller/TasksApiController.java
```diff
@@ -83,6 +88,18 @@ public EventEngineTaskDTO createTask(
...
   }
 
+  @PutMapping("/tenant/{tenantId}/tasks/{uuid}")
+  @ApiOperation(value = "Update specific Task for Tenant")
+  public EventEngineTaskDTO updateTask(
+      @PathVariable String tenantId,
+      @PathVariable UUID uuid,
+      @RequestBody @Validated(ValidationGroups.Update.class) CreateTask task
+  ) {
+    final EventEngineTask eventEngineTask = tasksService.updateTask(eventConversionService.convertFromInput(tenantId, uuid, task));
```
jjbuchan, line 98: I'm not sure why you're using `convertFromInput` instead of just `tasksService.updateTask(tenantId, uuid, task)`?

src/main/java/com/rackspace/salus/event/manage/services/TasksService.java
```diff
@@ -202,4 +150,114 @@ public void deleteAllTasksForTenant(String tenant) {
     eventEngineTaskRepository.findByTenantId(tenant, Pageable.unpaged())
         .forEach(task -> deleteTask(tenant, task.getId()));
   }
+
+  @Transactional
+  public EventEngineTask updateTask(@Valid EventEngineTask update) {
```
jjbuchan, line 155: This also relates to my comment below regarding the use of `convertFromInput`, as I don't think `@Valid EventEngineTask` is ideal here with how you've got things set up.

e.g. `CreateTask` could contain a `null` measurement field, which means `update` would now have a `null` measurement which is invalid.  I don't think `@Valid` will actually fail since the null setting is in the column annotation instead of `@NotNull` but the intent is the same.

(@itzg this always brings up a point of maybe we should be using the validation annotations in addition to column nullables on our entities)

src/main/java/com/rackspace/salus/event/manage/services/TasksService.java
```diff
@@ -202,4 +150,114 @@ public void deleteAllTasksForTenant(String tenant) {
...
+    EventEngineTask eventEngineTask = eventEngineTaskRepository.findByTenantIdAndId(update.getTenantId(), update.getId()).orElseThrow(() ->
+        new NotFoundException(String.format("No Event found for %s on tenant %s",
+            update.getId(), update.getTenantId())));
+    log.info("Updating event engine task={} with new values={}", update.getId(), update);
+    boolean needsUpdate = false;
+    if(!StringUtils.isEmpty(update.getName()) && !eventEngineTask.getName().equals(update.getName()))  {
+      log.info("changing name={} to {} ",eventEngineTask.getName(), update.getName());
```
jjbuchan, line 162: In the spirit of the logging conversations we've had, using this may help us once we have a better overall kibana solution.  There's more places below where we can do similar

```suggestion
      log.info("changing name={} to updatedName={} ",eventEngineTask.getName(), update.getName());
```

I like that you're adding these logs; we don't have enough and I think `info` is a valid level for these updates.

src/test/java/com/rackspace/salus/event/manage/web/controller/TasksApiControllerTest.java
```diff
@@ -258,6 +263,31 @@ public void testCreateTask() throws Exception {
...
+
+    CreateTask create = buildCreateTask(true);
+    EventEngineTask eventEngineTask = eventConversionService.convertFromInput(tenantId, uuid, create);
+
+    mockMvc.perform(put("/api/tenant/{tenantId}/tasks/{uuid}", tenantId, uuid)
+        .content(objectMapper.writeValueAsString(create))
+        .contentType(MediaType.APPLICATION_JSON)
+        .characterEncoding(StandardCharsets.UTF_8.name()))
+        .andDo(print())
+        .andExpect(status().is2xxSuccessful())
```
jjbuchan, line 283: To be more precise.

```suggestion
        .andExpect(status().isOk())
```

src/test/java/com/rackspace/salus/event/manage/services/TasksServiceTest.java
```diff
@@ -545,4 +557,133 @@ private static CreateTask buildCreateTask() {
...
+
+    final KapacitorTaskId taskId = new KapacitorTaskId()
+        .setBaseId(UUID.fromString("00000000-0000-0000-0000-000000000001"))
+        .setKapacitorTaskId("k-1");
+
+    eventEngineTask.setKapacitorTaskId(taskId.getKapacitorTaskId());
+    eventEngineTask.setId(taskId.getBaseId());
+
+    entityManager.persist(eventEngineTask);
+    entityManager.flush();
```
jjbuchan, line 574: You might be able to do
```
eventEngineTaskRepository.save(eventEngineTask);
```
instead of this.  I'd try that first to avoid using `entityManager` directly.

src/main/java/com/rackspace/salus/event/manage/services/TasksService.java
itzg: @jjbuchan yeah, now that @rohitsngh27 [figured out how to intercept and provide better error reporting on JPA constraint violations](https://github.com/racker/salus-telemetry-monitor-management/pull/197/files#diff-03f139be4207edb3b9d87c275c560960R68-R77) then I agree that we could now add more of those declarations to the entities.

src/main/java/com/rackspace/salus/event/manage/services/TasksService.java
```diff
@@ -152,34 +151,34 @@ public void deleteAllTasksForTenant(String tenant) {
   }
 
   @Transactional
-  public EventEngineTask updateTask(@Valid EventEngineTask update) {
-    EventEngineTask eventEngineTask = eventEngineTaskRepository.findByTenantIdAndId(update.getTenantId(), update.getId()).orElseThrow(() ->
+  public EventEngineTask updateTask(String tenantId, UUID uuid, CreateTask createTask) {
```
itzg, line 154: Since the method name of `updateTask` now conflicts with the argument name `createTask`, I'm wondering if `CreateTask` (and the argument name) should be renamed to something like `TaskCU` (to be consistent with `MonitorCU`) or `CreateUpdateTask` or something else.

src/main/java/com/rackspace/salus/event/manage/services/TasksService.java
```diff
@@ -152,34 +151,34 @@ public void deleteAllTasksForTenant(String tenant) {
...
 
     if(needsUpdate) {
-      eventEngineTask = changeMeasurementAndTaskParameters(eventEngineTask, update);
+      eventEngineTask = changeMeasurementAndTaskParameters(eventEngineTask, createTask);
     }
     return eventEngineTaskRepository.save(eventEngineTask);
   }
 
-  private EventEngineTask changeMeasurementAndTaskParameters(EventEngineTask eventEngineTask, EventEngineTask update) {
+  private EventEngineTask changeMeasurementAndTaskParameters(EventEngineTask eventEngineTask, CreateTask createTask) {
```
itzg, line 181: It looks like the second argument isn't used since the name change didn't cause any code changes. IntelliJ should be giving you a warning about the unused argument.

With that said, it feels like there was a reason in a previous PR for adding the argument, but perhaps double check it isn't needed.

itzg:
> In general it is best to allow anyone who has previously requested changes the chance to review again before you merge. Reviews are often a multi-person effort where Geoff might focus on some parts while i focus on other changes, but by combining them all we end up with cleaner code to maintain.

Agreed. I was expecting github to block the merge since the review had been marked as "requested changes".

PR #67, opened 2020-08-31
Updated tools execution and made sure to update upload to google cloud

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-990

# What

Make sure that we are generating the documentation in a way that google cloud storage will actually display the HTML.

# How

Simplifies the generation process by removing the mustache templates, generates the html with a different tool that isn't downloading js from some other site and uploads each of the public and admin html files on just a deploy. 

# How to test

Delete your entire targets directory then run the maven `package` command (after having updated and `mvn deploy`ed the new tools changes [linked at the bottom])

# Why

The biggest problem from this side was that we were generating multiple swagger html documents that needed to be uploaded and the tools used for that wouldn't generate each one with different names. Because of that the artifacts approach in the cloudbuilds file wouldn't work because it wanted to upload every file to one directory and uploading multiple files of the same name just overwrote that file. 

Ultimately using gsutils allowed us to upload each file to its own location without having to worry about what the filename was in the first place. 

# Depends on

https://github.com/racker/salus-tools/pull/41


cloudbuild-deploy.yaml
```diff
@@ -110,6 +110,10 @@ steps:
     - 'CLOUDSDK_COMPUTE_REGION=${_CLOUDSDK_COMPUTE_REGION}'
     - 'CLOUDSDK_CONTAINER_CLUSTER=${_CLOUDSDK_CONTAINER_CLUSTER}'
     - 'TILLERLESS=true'
+  - name: 'gcr.io/google.com/cloudsdktool/cloud-sdk'
+    args: ['gsutil', 'cp', 'target/generated/swagger/public/index.html', 'gs://salus_www/monitor_management/public']
```
GeorgeJahad, line 114: what is "gs://salus_www/monitor_management/public"

GeorgeJahad: should it be event-engine?

harageth: It should be. I created a salus_www bucket to place all of our static html and forgot to change each of these.

PR #73, opened 2020-10-20
Remove kapacitor as event engine

jjbuchan:
# Related PRs

https://github.com/racker/salus-telemetry-model/pull/162

# What
Removes the use of kapacitor ids in event engine tasks.

This PR does not touch the test-task feature.  It is just enough to begin modifying tasks to use esper instead of kapacitor.

## How to test

The majority of tests have been stripped out in this PR since they're no longer relevant.  New ones will be added in subsequent PRs to test new functionality.

# Why

Part of the process of switching from Kapacitor to Esper for the event engine.

# TODO

I'm going to create a chain of additional PRs to this one to fully build out the new model while trying to keep each individual one easy(ish) to reivew

PR #74, opened 2020-10-20
Abstract EventTask to allow subtypes for different monitoring systems

jjbuchan:
# Related PRs

https://github.com/racker/salus-event-engine-management/pull/73
https://github.com/racker/salus-telemetry-model/pull/162
https://github.com/racker/salus-telemetry-model/pull/163

# What

1. Allows for different types of tasks to be created for each monitor system
    * Uses changes from https://github.com/racker/salus-telemetry-model/pull/163
1. Had to also make DTOs and CU objects for each of those as well
   * We'll require different methods for handling all the different subtypes (currently only 2)
1. Made tests all use the generic type for now as it required the least change

# Note

* Uses `monitorScope` as the name for `ConfigSelectorScope`
    * This is also a proposal to change that in more places
    * It would make a little more sense to customers than `selectorScope`?
    * Anyone got a better name?

# Why

Each monitoring system may require different fields to be used for partitioning/filtering etc.

For salus the metrics will be partitioned by `tenantId:monitorType:monitorScope` and esper will also filter metrics using those fields.  That means those fields must be required in the task object.

However, maybe UIM doesn't have those fields and Tasks are configured differently.  For now it will fall under the `generic` subtype which can only use the `measurement` name.

This change allows us to add Salus specific changes that we know are needed without necessarily blocking us from handling other systems later on.

# TODO

Basic tests can probably be added for some of the new functionality.

I'm going to create a chain of additional PRs to this one to fully build out the new model while trying to keep each individual one easy(ish) to reivew

### DIFF
--- a/src/main/java/com/rackspace/salus/event/manage/services/TaskGenerator.java
+++ b/src/main/java/com/rackspace/salus/event/manage/services/TaskGenerator.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2020 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rackspace.salus.event.manage.services;
+
+import com.rackspace.monplat.protocol.UniversalMetricFrame.MonitoringSystem;
+import com.rackspace.salus.event.manage.model.GenericTaskCU;
+import com.rackspace.salus.event.manage.model.SalusTaskCU;
+import com.rackspace.salus.event.manage.model.TaskCU;
+import com.rackspace.salus.event.manage.web.model.EventEngineTaskDTO;
+import com.rackspace.salus.event.manage.web.model.GenericEventEngineTaskDTO;
+import com.rackspace.salus.event.manage.web.model.SalusEventEngineTaskDTO;
+import com.rackspace.salus.telemetry.entities.EventEngineTask;
+import com.rackspace.salus.telemetry.entities.subtype.GenericEventEngineTask;
+import com.rackspace.salus.telemetry.entities.subtype.SalusEventEngineTask;
+import org.springframework.stereotype.Component;
+
+@Component
+public class TaskGenerator {
+
+  public TaskGenerator() {
+  }
+
+  public EventEngineTask createTask(String tenantId, TaskCU in) {
+    EventEngineTask task;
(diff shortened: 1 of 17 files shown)


src/main/java/com/rackspace/salus/event/manage/services/TaskGenerator.java
```diff
@@ -0,0 +1,73 @@
...
+
+@Component
+public class TaskGenerator {
+
+  public TaskGenerator() {
+  }
+
+  public EventEngineTask createTask(String tenantId, TaskCU in) {
+    EventEngineTask task;
+    if (in instanceof SalusTaskCU) {
```
GeorgeJahad, line 39: I think there is a reason why we don't do this polymorphically, but I can't remember.  What was it again?

jjbuchan: I don't recall anything.  Maybe I'll see issues once I start adding tests?

itzg:
> Uses monitorScope as the name for ConfigSelectorScope

Yeah, I like that term.

src/main/java/com/rackspace/salus/event/manage/services/TasksService.java
```diff
@@ -140,4 +146,70 @@ public EventEngineTask updateTask(String tenantId, UUID uuid, TaskCU taskCU) {
...
+   * @param eventEngineTask
+   * @param taskCU
+   * @return True if any value was modified, otherwise false.
+   */
+  private boolean handleSystemSpecificTaskUpdate(TaskCU taskCU, EventEngineTask eventEngineTask) {
+    if (eventEngineTask instanceof SalusEventEngineTask && taskCU instanceof SalusTaskCU) {
+      return handleSalusTaskUpdate(taskCU, eventEngineTask);
+    } else {
+      return handleGenericTaskUpdate(taskCU, eventEngineTask);
+    }
```
itzg, line 162: I was recently reading about how the new sealed classes feature will help these if-instanceof-else scenarios

https://openjdk.java.net/jeps/360

itzg (approved):

src/main/java/com/rackspace/salus/event/manage/services/TasksService.java
jjbuchan: That is a nice improvement

src/main/java/com/rackspace/salus/event/manage/services/TaskGenerator.java
jjbuchan: I've not seen any problems.

GeorgeJahad (dismissed):
@jjbuchan writes:
> Basic tests can probably be added for some of the new functionality.

I would say they should be added and a ticket should be created so we don't forget.

jjbuchan:
@GeorgeJahad that was an old todo.  all tests are in https://github.com/racker/salus-event-engine-management/pull/75/files

jjbuchan:
@GeorgeJahad I pushed one more commit to here that fixed the null monitoringSystem errors we saw in the API.

The explanation:

>  Note on visibility of type identifier: by default, deserialization (use during reading of JSON) of type identifier is completely handled by Jackson, and is not passed to deserializers. However, if so desired, it is possible to define property visible = true in which case property will be passed as-is to deserializers (and set via setter or field) on deserialization.

PR #75, opened 2020-10-20
Add partition to event tasks

jjbuchan:
# Depends On

https://github.com/Rackspace-Segment-Support/umb-protocol/pull/9

# What

Adds a `partitionId` to all event tasks.

# How

Generates the id on save using the new custom kafka partitioner in umb.

# Why

The event-processor will only load tasks based on the kafka partitions it is consuming metrics on.

# TODO

Once this and the other 2 related event-mgmt PRs are all merged the main things leftover are to alter the test-task code.  There is another jira for that.

---

Before merging devops must first truncate the event tasks table - https://jira.rax.io/browse/SALUS-1059
e2e tests should not run during this time otherwise it could insert the old data format that we don't want before the new way has had time to deploy & startup.


src/test/java/com/rackspace/salus/event/manage/services/TaskPartitionIdGeneratorTest.java
```diff
@@ -0,0 +1,150 @@
...
+  KafkaTopicProperties kafkaTopicProperties;
+
+  @Test
+  public void testGetPartitionForTask_genericTask() {
+    InvalidRecordException exception;
+    GenericEventEngineTask eventEngineTask = (GenericEventEngineTask) new GenericEventEngineTask()
+        .setMeasurement("myMeasurement")
+        .setId(UUID.randomUUID())
+        .setName("myName")
+        .setTenantId("t-2000")
```
itzg, line 57: I went to double check which model was the T-2000 and I happened to see somebody added some funny hacks to this wiki page:

https://terminatorfanon.fandom.com/wiki/T-2000?oldid=5574

itzg (dismissed):

src/test/java/com/rackspace/salus/event/manage/services/TasksServiceTest.java
```diff
@@ -60,54 +78,266 @@
...
+    assertThat(retrieved).hasSize(2);
+
+    tasksService.deleteAllTasksForTenant("t-1");
+
+    retrieved = eventEngineTaskRepository.findByTenantId("t-1", Pageable.unpaged());
+    assertThat(retrieved).isEmpty();
+  }
+
+  @Test
+  @Transactional
```
GeorgeJahad, line 200: why transactional?

jjbuchan: I didn't dig into it but the previous test this was copied from was set as transactional.

GeorgeJahad (dismissed):

PR #76, opened 2020-10-27
 Update to use new MonitoringSystem enum & Fix API (de)serialization

jjbuchan:
Updating services to use the new `MonitoringSystem` enum in salus-models instead of the one in umb-protocol.

See https://github.com/racker/salus-telemetry-model/pull/165 for more info

---

This also fixes up the API usage problem that was leftover from the recently merged PRs,

### DIFF
--- a/src/main/java/com/rackspace/salus/event/manage/model/TaskCU.java
+++ b/src/main/java/com/rackspace/salus/event/manage/model/TaskCU.java
@@ -18,7 +18,7 @@ package com.rackspace.salus.event.manage.model;
 
 import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
-import com.rackspace.monplat.protocol.UniversalMetricFrame.MonitoringSystem;
+import com.rackspace.salus.telemetry.model.MonitoringSystem;
 import com.rackspace.salus.event.manage.model.ValidationGroups.Create;
 import com.rackspace.salus.event.manage.model.ValidationGroups.Test;
 import com.rackspace.salus.telemetry.entities.EventEngineTaskParameters;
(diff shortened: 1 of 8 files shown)


src/main/java/com/rackspace/salus/event/manage/model/TaskCU.java
```diff
@@ -29,8 +30,9 @@
 
 
 @Data
-@JsonTypeInfo(use = JsonTypeInfo.Id.NAME,
-    property = "monitoringSystem", visible = true, defaultImpl = GenericTaskCU.class)
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = As.EXISTING_PROPERTY,
+    property = "monitoringSystem", defaultImpl = GenericTaskCU.class,
+    visible = true)
```
jjbuchan, line 35: `visible = true` was added in a late commit before merging the previous PRs.  It allowed Task API calls to work.  Before that the service was always seeing the `monitoringSystem` value as null.

I expected it to break some tests, but it ended up breaking them in more ways than I expected.  For some reason the deserialization acted a little differently in tests than it appeared to when running the service and making an api call myself.

By adding `include = As.EXISTING_PROPERTY` it tells the serialization to use property defined on the object and not use the new one generated by `@JsonTypeInfo`.  With this `include` option we now get `{"monitoringSystem" : "UIM"}`.  Without it I was seeing two `monitoringSystem` values in the test payload - `{"monitoringSystem" : "UIM", "monitoringSystem": "GenericTaskCU"}`

I believe `visible = true` is needed here but not for Monitor creation is because an enum is being used instead of a string.  The `include` makes sense as shown by the default below.  For monitors `type` in not a regular property, it is only defined by the `JsonTypeInfo` annotation.

>JsonTypeInfo.As
>
> EXISTING_PROPERTY
> Inclusion mechanism similar to PROPERTY with respect to deserialization; but one that is produced by a "regular" accessible property during serialization.
>
> PROPERTY (default)
> Inclusion mechanism that uses a single configurable property, included along with actual data (POJO properties) as a separate meta-property.

itzg: The new change makes sense, but wow, that was a subtle one to figure out.

itzg (approved):

PR #81, opened 2020-11-03
Send TaskChangeEvents on all write operations

jjbuchan:
# What

Send `TaskChangeEvent`s to kafka for any create, update, delete.

# How to test

Updated existing tests to additionally check for the new producer operations.

# Why

Event-Engine-Processor will read these and (un)deploy the corresponding queries to esper.

### DIFF
--- a/src/main/java/com/rackspace/salus/event/manage/services/TaskEventProducer.java
+++ b/src/main/java/com/rackspace/salus/event/manage/services/TaskEventProducer.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2020 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


src/main/java/com/rackspace/salus/event/manage/services/TaskEventProducer.java
```diff
@@ -0,0 +1,56 @@
...
+    this.kafkaTemplate = kafkaTemplate;
+    this.properties = properties;
+  }
+
+  public void sendTaskChangeEvent(TaskChangeEvent event) {
+    final String topic = properties.getTaskChanges();
+
+    log.debug("Sending taskChangeEvent={} on topic={}", event, topic);
+    try {
+      kafkaTemplate.send(topic, buildMessageKey(event), event).get();
```
jjbuchan, line 50: One thing I noticed while doing this is that in some places we have the blocking `get()` operation on event sends and in other places we don't.

I'm assuming the asynchronous benefit doesn't really matter for us in these cases and it's better to catch and log the error than having it fail silently (unless we add on additional async handling of those errors).

Just want to make sure we're making a conscious decision here.

If we did this in the ambassador where we send metrics I'd be more concerned, but we don't.

itzg (approved):

GeorgeJahad (approved):

GeorgeJahad:
I've updated the e2e for the api changes but the events aren't being generated, mainly i think because this pr hasn't been merge d yet.  other than this one, are there any other prs required before the e2e should start working?

